PR #23, opened 2020-07-15
Feature/add GitHub actions

otahontas:
### Description

Adding github actions

### Type of change

### Testing

- [x] Tests do not apply


blankdots:
these don't seem to be triggered :thinking:

blankdots:
@otahontas see #24 for fixes.

PR #27, opened 2020-07-17
Feature/add json forms and GitHub actions

otahontas:
### Description

This PR adds github actions and basic jsons forms. Currently only study schema is supported and it's loaded from frontend. We'll fix this when schema querying is supported in backend.

### Type of change

### Changes Made

- Added github actions (or rather rebased gh action branch to this branch)
- Added form rendering based on json schemas, currently using study schema
- Added some new styles

### Testing


Dockerfile-dev
```diff
@@ -1,4 +1,4 @@
-FROM node:12-alpine
+FROM node:14-alpine
```
blankdots, line 1: might also need to update: https://github.com/CSCfi/metadata-submitter/blob/develop/Dockerfile#L1

otahontas: true

Issue #55, opened 2020-08-10
Fix layout bugs in forms


otahontas:
#### Describe the bug

In some forms, save button is in rights side and there are some weird layout issues (probably because of untested usage of flow and percentage css attributes). 

#### To Reproduce

Steps to reproduce the behaviour:

1. Go to sample json form
2. See weird save button positioning

#### Expected behavior

Form fields are in row-based order

PR #57, opened 2020-08-10
Feature/add oneOf rendering

### LINKED ISSUE #42: Fix OneOf -type rendering and validation
#### Describe the bug

Our form rendeder and Yup-schema builder dont' currently work well with oneOf -schema keywords. For this problem, we could follow react-jsonschema-forms approach where each option of OneOf -array are presented as select field, and after user has selected one option, that option is rendered. Look for "Policy URL / policy text" in policy schema for example

Following changes should be made:
- Optional form field rendering based on user select, see formik docs for dependent fields and array-based rendering
- Custom config to yup schema, where oneOf types are better recognized

#### To Reproduce

1. open fill form option for either run, experiment, analysis or policy
2. Form renders, but only partially. See console for errors.

#### Expected behavior

Form renders without problems

otahontas:
### Description

This PR adds support for oneOf-fields: user can select which option he/she wants to use and correct fields are rendering according to selection. Layout issues such as save-button positioning and headings for arrays are also fixed.

### Related issues
Fixes #42, #55

### Type of change

### Testing


otahontas:
> Noticed
> image
> 
> and made a PR with more typo fixes: [CSCfi/metadata-submitter#117](https://github.com/CSCfi/metadata-submitter/pull/117)

There are actually a few different keys without any titles, which should be fixed at some point. See for example https://github.com/CSCfi/metadata-submitter/blob/950555f0b7309fbd4379dc0bbfbf4f13ae96187d/metadata_backend/helpers/schemas/ena_experiment.json#L364

Currently version is to render path of field (such as `descriptor.studyAttributes`) as label is title is not found

Issue #77, opened 2020-09-02
Fix "draft version" alert from popping up on clicks where it shouldn't


csc-jm:
#### Describe the bug

Currently the alert, "Would you like to save draft version of this form?", pops up if a form is being filled and the user tries to switch to add another object. However, the buggy behavior is that the alert pops up even if nothing in the form has been filled yet. Moreover, the same alert appears if the user is in the "Upload XML file" card and tries to switch to add another object.

#### To Reproduce

Steps to reproduce the behaviour:

1. Click "Create Submission"
2. Click "New Folder"
3. Enter whatever as name and description and click next
4. Open any option from the left sidebar
5. Click "Fill form"
6. Click another option e.g. "Upload XML File"
7. Alert appears

Additionally:
8. Go to "Upload XML file"
9. Now click another option e.g. "Fill Form"
10. Alert appears

#### Expected behavior

The alert should probably only pop up when something has been entered in to the form, not when the form is still empty. In the latter example the alert should not appear at all.
When the form or upload field has changes, dispatch a state that informs about form/upload status. Clear this state if cancelled or saved from dialog

#### Screenshots

Alert in question:
![Screenshot 2020-09-02 at 11 00 41](https://user-images.githubusercontent.com/65757007/91955254-947f1100-ed0b-11ea-9644-46d7d296ce65.png)

saulipurhonen:
Side note: Formik has "touched" -property which checks if form has been edited. This could be useful in change detection.

saulipurhonen:
As we're using React Hook Form instead of Formik we need to check forms isDirty state, which compares current value to fields defaultValue. It seems that the form doesn't register default values just yet. I can see that WizardJSONSchemaParser introduces a defaultValue field in form fields but this isn't accessible in form. I'll keep investigating.

saulipurhonen:
This can be solved with formState.isDirty method.
You can log isDity status for example in WizardFillObjectDetailsForm:

**FormContent** function
console.log(methods.formState.isDirty)

lilachic:
Thanks. That leaves putting and checking that status to/from global state in draftStatus.

PR #81, opened 2020-10-07
Feature/consistent alerts

saulipurhonen:
### Description

Created component for alert dialog. Component renders dialog content based on parent component location and alert type.
NOTE: When #77 is done, refactor alerts to be called only when form/upload has changed

### Type of change

### Changes Made

- Moved dialog logic from object index and footer components to form alert component
- Added functionality for stepper to check if submission type is available and render alert based on that condition

### Testing


src/components/NewDraftWizard/WizardComponents/WizardFooter.js
```diff
@@ -112,8 +82,10 @@ const WizardFooter = () => {
               color="primary"
               disabled={wizardStep < 1}
               className={classes.footerButton}
+              // TODO: Implement save functionality
```
blankdots, line 85: I think we can create an issue out of it for better tracking if there isn't one yet.

blankdots (approved):
LGTM.

> NOTE: There needs to be a check if form / upload field values have changed from initial render.

is that a new issues or how do we track it

src/components/NewDraftWizard/WizardComponents/WizardFooter.js
saulipurhonen: > NOTE: There needs to be a check if form / upload field values have changed from initial render.

This is addressed in #77 

> I think we can create an issue out of it for better tracking if there isn't one yet.

Yeah, this is a different issue since save functionality is out of scope of dialog alerts.

Issue #93, opened 2020-11-10
Changing form submission between object types causes too many re-renders


saulipurhonen:
#### Describe the bug

If user moves between different object types form submission, you can see that the load times for forms are taking longer. This affects checking form status and can give false positives on isDirty. Might also be related to "clear form" action which doesn't seem to work right.

#### To Reproduce

Steps to reproduce the behaviour:

1. Go to WizardFillObjectDetailsForm component
2. In FormContent method log formschema
3. Move between forms on different object types
3. Console logs form schema multiple times

#### Expected behavior

Form should be rendered only once when object type changes.

saulipurhonen:
Seems to be related to form resolver.

saulipurhonen:
It seems that if user navigates to xml upload between different object type form, resolver is fired only once. This might be fixed with nulling form data between object type change.

saulipurhonen:
Fixed with destroying form component between changing object types. Form component is given a key (a random number at this point, should be more convenient). Component is destroyed when the key changes and therefore doesn't do unnecessary validations. This is going to be fixed when **bugfix/prevent-draft-version-alert-popping-incorrectly** is merged.

Fixed form reset also.

PR #95, opened 2020-11-11
Bugfix/prevent draft version alert popping incorrectly

### LINKED ISSUE #77: Fix "draft version" alert from popping up on clicks where it shouldn't

saulipurhonen:
### Description

Use draft status as state. Monitor this state when navigating between steps and object / submission types.

### Related issues

Fixes #77 and #93 

### Type of change

### Changes Made

- Monitor draft status and fire draft dialog if unsaved changes
- Destroy object card component after object / submission type change to prevent memory leaks
- Check if form is edited and set draft status
- Updated tests to match current code

### Testing

- [x] Tests do not apply

### Mentions

@lilachic initialized draft status state.


lilachic (approved):
Plenty of places to handle draftStatus.

blankdots (approved):

PR #97, opened 2020-11-13
Feature/folder backend check

### LINKED ISSUE #65: Add error checks for folder related backend operations
#### Proposed solution

Currently results from folder-related backend operations are not checked, so if backend is not running / request fails, wizard will just continue and crash later. 

Following situations are problematic currently:
- Addition of folders: if posting to backend fails, wizard continues to second step
- Addition of objects to folders: if adding to folder fails in backend, nothing is reported (however folder isn't added to redux store in frontend, which is good). As a result, object is in backend, but not connected to folder
- Deletion of folder: if deleting folder or objects inside it fails, nothing is reported (also folder is not removed from store at all if any of requests to backend fails).

We should handle error checks for backend operations and report fails to user +  in best situation these. 

#### DoD (Definition of Done)

Errors are checked and handled accordingly, user is notified

saulipurhonen:
### Description

Check folder API response and set error if bad response. This is done in wizardSubmissionFolderSlice methods where API calls are made.

### Related issues

Fixes #65 

### Type of change

### Changes Made

- Check against API call responses in promise
- Catch and pass errors to components
- Show error messages via status message handler in snackbar

### Testing

- [x] Needs testing

### DIFF
--- a/src/components/NewDraftWizard/WizardComponents/WizardAlert.js
+++ b/src/components/NewDraftWizard/WizardComponents/WizardAlert.js
@@ -7,9 +7,7 @@ import DialogActions from "@material-ui/core/DialogActions"
 import DialogContent from "@material-ui/core/DialogContent"
 import DialogContentText from "@material-ui/core/DialogContentText"
 import DialogTitle from "@material-ui/core/DialogTitle"
-import Link from "@material-ui/core/Link"
 import { useSelector } from "react-redux"
-import { Link as RouterLink } from "react-router-dom"
 
 /*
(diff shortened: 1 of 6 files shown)


saulipurhonen:
Example error reproduce:

1. Go into folder creation step of wizard
2. change baseurl to "/folders1" in services/folderAPI.js
3. Fill form and navigate to next step
4. Snackbar should appear

lilachic (approved):
Thanks for giving an example of how to reproduce errors .

blankdots (approved):

Issue #101, opened 2020-11-23
Handle additional folder draft operations


csc-jm:
#### Proposed solution

After refactoring the folders to include drafts in the backend in [#144](https://github.com/CSCfi/metadata-submitter/pull/144), the frontend needs to take this into account. This means, when drafts are created, the folder where the draft belongs in should be patched accordingly. Also for publishing a folder, there is a new endpoint `/publish/{folderId}`. This endpoint discards the unfinished drafts within the folder and deletes them from the draft database collection. Therefore, upon publishing the user should be alerted that the drafts will be discarded.

#### DoD (Definition of Done)

-Additional `PATCH` operations are added for adding and removing drafts in folders
-Alerts are added for when folders are being published and drafts within the folders will be removed

blankdots:
dependent also on https://github.com/CSCfi/metadata-submitter/issues/153

saulipurhonen:
Pending on #67

PR #112, opened 2020-12-11
Feature/auto save draft

### LINKED ISSUE #67: Implement (auto-)saving draft versions of folders
#### Proposed solution

If user currently changes away from form, alert is shown asking if user want's to save the form. Save doesn't do anything, so implement draft possibility. 

User should also be notified somehow that they have non-finished versions of forms available. 

In best case form is auto-saved every now and then and saving is reported to user via notification.

#### DoD (Definition of Done)

Saving draft versions is possible.

saulipurhonen:
### Description

Enables draft saving functionality via dialogs and by auto-saving with timer.
Folder publish is done through /publish endpoint. This call removes drafts from the folder.
Draft save notification and connection errors are shown through status message handler.

NOTE: To get authorization, you need to start from login page.

### Related issues

Closes #67 & #101

### Type of change

### Changes Made

- Login link through backend port to get authorization.
- Enabled draft save with dedicated function
- Start timer when form changes, timer resets on every key press
- Reset timer when dialog is open or draft is saved
- Slice for status message (status, response, prefix text)
- Slice for alert. Helps with stopping auto-save timer
- Show draft related status messages in wizard root
- Added "/publish" -endpoint into proxy config

### Testing

### Mentions

Cheers to @blankdots for helping with backend related problems.


saulipurhonen:
I introduced status message handler in wizard root component because we need to inform when draft is saved. This component is fired if statusDetails slice has updates. Status details are updated only when draft is saved but this same logic could be used in other places where status message component is called.
This would reduce need to have status message component only in wizard root.

blankdots (commented):
identified: https://github.com/CSCfi/metadata-submitter/pull/164 and fixed

blankdots (commented):
I am not sure where to recover the saved draft objects. I am expecting them to be in `Choose existing object`, but I assume this can be done in a different PR.

blankdots (approved):
functionality-wise LGTM.
would have liked a way to resume the draft, but i assume this can be done in a different PR

hannyle (approved):
I also think that we should be able to retrieve the draft objects somewhere.
Otherwise, it works fine for me.

csc-jm (approved):
Adding on to what others mentioned. I was initially confused as well that there doesn't seem to be a way to continue a previous draft yet but that functionality is big enough for another PR. However, I would like to note that in this PR, every auto-save creates a new draft object altogether which I feel is not the desired way. For example if you are filling a form and you fill one field, wait for 5 seconds, then fill another field and wait for another 5 seconds, two different drafts are created even though it would make more sense that the original draft was patched.

Everything else here seems to work as defined in the PR description though so I will also approve this as is.

lilachic (approved):
This is related to previous reviews, also without changing the same form is possible to be saved multiple times with save button. And here remains a problem with long forms (analysis e.g.), auto-save saves but the form lacks some required fields, making the work flow not so fluent.

lilachic:
And a question about the wizardAlertSlice why it uses the name `name: "draftStatus”, line 7 ?

blankdots:
@saulipurhonen we will move comments on this PR to issues might be good idea to also review those.

Issue #117, opened 2020-12-16
Catch errors 401, 403, 500 


hannyle:
#### Proposed solution

Catch errors 401, 403, 500 from back-end and direct them to error handling pages 401, 403, 500
Related to https://github.com/CSCfi/metadata-submitter-frontend/issues/96

#### DoD (Definition of Done)

Error handling pages 401, 403, 500 will show whenever those errors appear

#### Testing

Unit test and peer review

PR #119, opened 2020-12-23
Add apisauce monitor to check api requests and catch errors

hannyle:
### Description

Catch errors 401, 403, 500 from back-end and direct them to error handling pages 401, 403, 500

### Related issues

https://github.com/CSCfi/metadata-submitter-frontend/issues/117

### Type of change

### Changes Made

- New file `errorMonitor.js` created to handle the redirection when errors 401, 403, 500 occur
- `apisauce monitors` are added to `/services` api calls

### Testing

### Mentions

- I tried to come up with a suitable unit test for the `monitors` but haven't figured out any good ways so far. Not sure if we need unit test for this.
- Some errors cannot be caught from client-side, may need to do the redirection from server-side (to be implemented later)

### DIFF
--- a/src/services/errorMonitor.js
+++ b/src/services/errorMonitor.js
@@ -1,0 +1,18 @@
+// function to catch errors
+export const errorMonitor = res => {
+  if (!res.ok) {
+    switch (res.status) {
+      case 401:
+        window.location = "/error401"
+        break
+      case 403:
+        window.location = "/error403"
+        break
+      case 500:
+        window.location = "/error500"
+        break
+      default:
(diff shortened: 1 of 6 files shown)


blankdots:
backend redirect to be handled in this issue: https://github.com/CSCfi/metadata-submitter/issues/170

blankdots (commented):
Changes look good, will test it with back-end and then come back with approval or suggestions.

blankdots (approved):
once the backend issue is also solved we can come back and do some integration tests for this

Issue #107, opened 2020-12-02
Draft Objects should be notified


hannyle:
#### Description

There should be a notification of the amount of draft versions of each object.

#### DoD (Definition of Done)

Have number (for e.g. `1`) next to the title of an object (According to the design) to notify that there is a draft version of an object 

![Screenshot 2020-12-28 at 18 28 50](https://user-images.githubusercontent.com/23532204/103228909-8c8e5100-493a-11eb-9ed4-0b798739b502.png)

#### Testing

Peer review

blankdots:
Dependent on #67

blankdots:
we will need to follow up this with #116 in order to see how to use those drafts

Issue #116, opened 2020-12-15
see saved drafts in a folder


blankdots:
#### Description

As a follow-up of #112  we would like to see drafts for each type of object and reuse them in the wizard

we can also rename `Choose existing object` to `Choose from drafts`

The drafts should be saved at longer intervals e.g. 60s

Add also a button to remove/delete drafts

#### DoD (Definition of Done)

drafts are visible in `Choose existing object`

#### Testing

unit testing, and if possible integration tests

blankdots:
also take into consideration not to overwrite data as mentioned here: https://github.com/CSCfi/metadata-submitter-frontend/pull/112#pullrequestreview-553544444

PR #121, opened 2020-12-30
Feature/notify draft object

hannyle:
### Description
Each object would have a notification next to its title when there is any drafts been saved

### Related issues

https://github.com/CSCfi/metadata-submitter-frontend/issues/107

### Type of change

### Changes Made

- Add a notification badge next to object title to count the saved drafts in each object
- Change button `Save` to `Submit object`
- Auto-saving time for draft changes from 5s to 60s
- New button `Save as Draft` is added so the user can manually save a form as a draft object
- Unit test for showing the correct number of saved drafts in each object

### Testing

- [x] Unit Tests

### DIFF
--- a/cypress/integration/app.spec.js
+++ b/cypress/integration/app.spec.js
@@ -17,12 +17,12 @@ describe("Basic e2e", function () {
     cy.get("textarea[name='description']").type("Test description")
     cy.get("button[type=button]").contains("Next").click()
 
-    // Fill a study form and save object
+    // Fill a study form and submit object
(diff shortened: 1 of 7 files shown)


csc-jm (approved):
Changes seem to work so looks good to me. Still unsure how the number of drafts will play into using them but I guess that's an issue for another PR.

blankdots:
Looks good to me as well pretty sure we need the #116 to clarify how we can use them.

Will merge as is

Issue #31, opened 2020-07-22
Add tests for wizard layout


otahontas:
#### Proposed solution

Wizard process isn't tested in any way currently, add some basic unit tests, later end-to-end

#### DoD (Definition of Done)

Wizard is extensively tested

#### Testing

Part of solution

saulipurhonen:
I think this can be closed since we are testing UI in both integration and e2e tests. 
@blankdots, any opinions?

blankdots:
I think for this to be fully complete we need to retrieve the folder that has been submitted in the home page listing. But most likely we probably need a different test for that.

However yes this could be considered fixed with #108

blankdots:
this was already tested in #108 but we can add as part of this testing with min 2 forms/objects (e.g. study and analysis) being filled in and added to the submission as an extension of the end 2 end tests.
it should also check summary that it contains the right amount of objects that have been saved.

PR #125, opened 2021-01-08
Add test for Analysis object Fill form and check Summary amount of ob…

hannyle:
### Description

Add test for filling form in Analysis object and check the Summary of objects if it has the correct amount

### Related issues

https://github.com/CSCfi/metadata-submitter-frontend/issues/31

### Changes Made

- Extend `Wizard Layout` test in `app.spec.js`: Filling form in Analysis object and submit it
- Test that Summary has the correct number of Study objects and Analysis object

### Testing

- [x] e2e test

### DIFF
--- a/cypress/integration/app.spec.js
+++ b/cypress/integration/app.spec.js
@@ -45,7 +45,10 @@ describe("Basic e2e", function () {
 
     // Fill an Analysis form and submit object
     cy.get("div[role=button]").contains("Analysis").click()
-    cy.get("div[role=button]").contains("Fill Form").click()
+    cy.get("div[role=button]")
+      .contains("Fill Form")
+      .should("be.visible")
+      .then($btn => $btn.click())
 
     cy.get("form").within(() => {
       // Experiment
--- a/cypress/integration/app.spec.js
+++ b/cypress/integration/app.spec.js
@@ -6,7 +6,7 @@ describe("Basic e2e", function () {
     cy.get('[alt="CSC Login"]').click()
   })
 
-  it("should create new folder, add study form and publish folder", () => {
+  it("should create new folder, add Study form, upload Study XML file, add Analysis form and publish folder", () => {
     cy.visit(baseUrl)
     cy.get('[alt="CSC Login"]').click()
     cy.visit(baseUrl + "newdraft")
@@ -44,12 +44,8 @@ describe("Basic e2e", function () {
     cy.get(".MuiListItem-container", { timeout: 10000 }).should("have.length", 2)
 
     // Fill an Analysis form and submit object
-    cy.get("div[role=button]")
-      .contains("Analysis")
-      .then($analysis => {
-        $analysis.click()
-        cy.get("div").contains("Fill Form").click()
-      })
+    cy.get("div[role=button]").contains("Analysis").click()
+    cy.get("div[role=button]").contains("Fill Form").click()
 
     cy.get("form").within(() => {
       // Experiment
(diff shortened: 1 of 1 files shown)


cypress/integration/app.spec.js
```diff
@@ -40,11 +40,87 @@ describe("Basic e2e", function () {
...
     cy.get("form").submit()
 
-    // Saved objects list should have newly added item
+    // Saved objects list should have newly added item from Study object
     cy.get(".MuiListItem-container", { timeout: 10000 }).should("have.length", 2)
 
-    // // Navigate to summary and publish
+    // Fill an Analysis form and submit object
+    cy.get("div[role=button]").contains("Analysis").click()
+    cy.get("div[role=button]").contains("Fill Form").click()
```
blankdots, line 48:
```suggestion
    cy.get("div[role=button]").contains("Fill Form").click({force: true})
```

I am getting the error below for some reason:

image

blankdots (changes requested):
for me it fails locally due to that error could we figure out another way to click that or use `{force: true}` ?

cypress/integration/app.spec.js
saulipurhonen: There seems to be latency in Cypress and for some reason it sees the list as hidden too long. One option would be to add longer timeout as parameter for `get` method.

As an side note:
I've seen Cypress to pass if ran again in headless browser. Initial run can expose some problems with Cypress latency issues but these need to be addressed since these issues may exist with automated e2e tests on GH.

hannyle: Could you try with my latest fix

cypress/integration/app.spec.js
```diff
@@ -40,11 +40,90 @@ describe("Basic e2e", function () {
...
+    // Fill an Analysis form and submit object
+    cy.get("div[role=button]").contains("Analysis").click()
+    cy.get("div[role=button]")
+      .contains("Fill Form")
+      .should("be.visible")
+      .then($btn => $btn.click())
+
+    cy.get("form").within(() => {
+      // Experiment
+      cy.get("input[name='experimentRef.accessionId']").type("Experiement Test Accession Id")
```
blankdots, line 55:
```suggestion
      cy.get("input[name='experimentRef.accessionId']").type("Experiment Test Accession Id")
```

cypress/integration/app.spec.js
```diff
@@ -40,11 +40,90 @@ describe("Basic e2e", function () {
...
+      cy.get("input[name='experimentRef.identifiers.submitterId.namespace']").type("Experiement Test Namespace")
```
blankdots, line 56:
```suggestion
      cy.get("input[name='experimentRef.identifiers.submitterId.namespace']").type("Experiment Test Namespace")
```

cypress/integration/app.spec.js
```diff
@@ -40,11 +40,90 @@ describe("Basic e2e", function () {
...
+      cy.get("input[name='experimentRef.identifiers.submitterId.value']").type("Experiement Test Value")
```
blankdots, line 57:
```suggestion
      cy.get("input[name='experimentRef.identifiers.submitterId.value']").type("Experiment Test Value")
```

blankdots (approved):
small typos otherwise works fine. Approved!

cypress/integration/app.spec.js
blankdots: works now. thx!

lilachic (approved):
Works for me.

Issue #113, opened 2020-12-11
Test draft saving


saulipurhonen:
#### Proposed solution

Draft (auto) saving isn't tested. Write tests to ensure saving works.

#### DoD (Definition of Done)

Tests run without problems.

PR #126, opened 2021-01-11
Feature/saved drafts in folder

### LINKED ISSUE #110: Store schemas in sessionStorage instead of localStorage
#### Proposed solution

If schemas update often, some users might not get the latest changes if schemas are stored in localStorage. Store form schemas into sessionStorage instead.

#### DoD (Definition of Done)

App stores form schemas into sessionStorage.

#### Testing

Check browser storage.

saulipurhonen:
### Description

Component for listing drafts with edit and delete functionality. Introduced logic for overwriting draft so (auto)saving current draft won't produce new draft.

Use session storage instead of local storage for setting schemas into browser storage.

Enable folder name and description patch to database.

### Related issues

Closes #110 #113 #116 

### Type of change

### Changes Made

- Renamed `Choose existing object` to `Choose from drafts`
- List drafts with edit and delete functions
- Added 401 error message case to  `WizardStatusMessageHandler`
- Delete draft when object is submitted
- Use session storage instead of local storage
- Patch folder name & description updates to backend
- Added unit test
- Added e2e tests
- Reset auto-save timer when form is submitted

### Testing

- [x] Unit Tests
- [x] E2E tests


blankdots (changes requested):
1. I think this needs a rebase
2. I think we should split the big integration test in `app.spec.js` into multiple smaller ones, preferably in different files, some that deal with drafts or leave in same file but do a different scenario for drafts e.g. `A user creates a draft and continues it, and publishes a form with it`

blankdots (commented):
image
Instead of the `X` i would put a 2 buttons `continue` and `delete` to make the actions explicit Wondering if https://material-ui.com/components/button-group/#button-group can have different colours `blue` for continue and  `red` for `delete` :thinking: if not we can have 2 separate buttons

hannyle:
Regarding to error 401, we have this file `services/errorMonitor.js` to handle API errors catches from client side, since `services/draftAPI` was added later so it hasn't had this yet.

Import and add the line `api.addMonitor(errorMonitor)` in `services/draftAPI` should solve the issue. I don't think we need to catch 401 in `WizardStatusMessageHandler.js`.
```suggestion
import { errorMonitor } from "./errorMonitor"
const api = create({ baseURL: "/drafts" })
api.addMonitor(errorMonitor)
```

hannyle (commented):
For me it's good that now the same drafts are saved in the same version, and submit the draft would delete it.
One thing I think we can improve (either in this PR or another PR) is that:
- if I open a draft from `Choose from drafts`, it opens a form with pre-filled values (which is correct)
- However if I change my mind and don't want to continue with this draft, I want to cancel the editing and have a new empty form, I need to click another `submission type` in `WizardObjectIndex` on the left, then click `Fill form` again. It seems that I have no way else to cancel from editing the current draft form.
- The same issue happens if I submit a draft. After I click Submit, the values of submitted form still exists, I need to do the same step above to cancel and have a new form.

blankdots:
@hannyle to your points:

>     * However if I change my mind and don't want to continue with this draft, I want to cancel the editing and have a new empty form, I need to click another `submission type` in `WizardObjectIndex` on the left, then click `Fill form` again. It seems that I have no way else to cancel from editing the current draft form ?
> 
>     * The same issue happens if I submit a draft. After I click Submit, the values of submitted form still exists, I need to do the same step above to cancel and have a new form.

I think we should have another issue for them, it sounds too complicated to include in this PR.

lilachic (approved):

saulipurhonen:
> 1. I think this needs a rebase
> 2. I think we should split the big integration test in `app.spec.js` into multiple smaller ones, preferably in different files, a file that deals with drafts and a file that deals with a submission without drafts, or leave in same file, but do a different scenario for drafts e.g. `A user creates a draft and continues it, and publishes a form with it`

Branch is rebased and E2E tests are separated.

saulipurhonen:
> image
> Instead of the `X` i would put a 2 buttons `continue` and `delete` to make the actions explicit.
> Wondering if https://material-ui.com/components/button-group/#button-group can have different colours `blue` for continue and `red` for `delete` 🤔 if not we can have 2 separate buttons

Listing drafts uses button groups with different color for each action.

saulipurhonen:
> For me it's good that now the same drafts are saved in the same version, and submit the draft would delete it.
> One thing I think we can improve (either in this PR or another PR) is that:
> 
> * if I open a draft from `Choose from drafts`, it opens a form with pre-filled values (which is correct)
> * However if I change my mind and don't want to continue with this draft, I want to cancel the editing and have a new empty form, I need to click another `submission type` in `WizardObjectIndex` on the left, then click `Fill form` again. It seems that I have no way else to cancel from editing the current draft form ?
> * The same issue happens if I submit a draft. After I click Submit, the values of submitted form still exists, I need to do the same step above to cancel and have a new form.

@hannyle, these are good points. Can you create a new issue for these suggestions? We have `Clear form` button and I think we'd extend its functionality to clear draft.

saulipurhonen:
> Regarding to error 401, we have this file `services/errorMonitor.js` to handle API errors catches from client side, since `services/draftAPI` was added later so it hasn't had this monitor yet.
> 
> Import and add the line `api.addMonitor(errorMonitor)` in `services/draftAPI` should solve the issue. I don't think we need to catch 401 in `WizardStatusMessageHandler.js`.
> 
> ```
> import { errorMonitor } from "./errorMonitor"
> const api = create({ baseURL: "/drafts" })
> api.addMonitor(errorMonitor)
> ```

Thanks! This is implemented in this PR.

blankdots (approved):

hannyle (approved):

Issue #127, opened 2021-01-11
Form isDirty status clear


saulipurhonen:
#### Describe the bug

Form isDirty status is set to true even after the form is cleared. Status is set to true after focusing any form fields.

#### To Reproduce

Steps to reproduce the behaviour:

1. Go to `Fill Form` submission type
2. Click on any form fields
3. Change submission type
4. Draft saving dialog opens

#### Expected behavior

Draft saving shouldn't be alerted when form is empty

saulipurhonen:
Form clear button doesn't clear `isDirty` status.

PR #146, opened 2021-02-01
Bugfix/clear form button

### LINKED ISSUE #143: Clear form does not seem to work
#### Describe the bug

clear form does not work

#### To Reproduce

1. Go to form
2. fill form
3. click clear form
4. does not work

#### Expected behavior

clear form

saulipurhonen:
### Description

Fixes issue where form doesn't clear when continuing draft.
Added check for form submit that form is valid.

### Related issues

Closes #143 

### Type of change

### Changes Made

- Reset form with form schema as parameter. This was with default values before and therefore wouldn't clear when editing draft
- Check that form is valid when submitting form. Without this current draft item would be deleted.

### Testing

- [x] E2E Tests

### DIFF
--- a/src/components/NewDraftWizard/WizardForms/WizardFillObjectDetailsForm.js
+++ b/src/components/NewDraftWizard/WizardForms/WizardFillObjectDetailsForm.js
@@ -96,7 +96,7 @@ const FormContent = ({ resolver, formSchema, onSubmit, objectType, folderId }: F
   const increment = useRef(null)
 
   const resetForm = () => {
-    methods.reset()
+    methods.reset(formSchema)
   }
 
   const checkDirty = () => {
@@ -256,7 +256,7 @@ const FormContent = ({ resolver, formSchema, onSubmit, objectType, folderId }: F
             className={classes.formButtonSubmit}
             onClick={() => {
               handleReset()
-              if (currentDraftId) handleDraftDelete(currentDraftId)
+              if (currentDraftId && methods.formState.isValid) handleDraftDelete(currentDraftId)
             }}
           >
             Submit {objectType}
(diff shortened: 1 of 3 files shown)


src/components/NewDraftWizard/WizardForms/WizardFillObjectDetailsForm.js
```diff
@@ -153,6 +153,7 @@ const FormContent = ({ resolver, formSchema, onSubmit, objectType, folderId }: F
     if (currentDraftId || draftObject.accessionId) {
       const response = await draftAPIService.patchFromJSON(objectType, currentDraftId, cleanedValues)
       if (response.ok) {
+        console.log(response)
```
blankdots, line 156: forgotten console log ?

saulipurhonen: Removed and rebased.

blankdots (approved):

lilachic (approved):
Looks good to me. Just a note that it clears a form but it still alerts asking to cancel/save, when trying to change to another object type.

blankdots:
@lilachic maybe that is another bug e.g. if the form is empty changing from it it should not raise that alert, or @saulipurhonen should we treat as part of this issue. I remember this was solved some time ago in #77 ?

saulipurhonen:
@lilachic, @blankdots that indeed is a bug and is issued in #127. I'll update the issue with more details regarding this behaviour.

hannyle:
looks good to me.

hannyle (approved):

blankdots:
merging as is and the identified bug will be fix in #127

Issue #135, opened 2021-01-18
Option to cancel editing a saved draft 


hannyle:
#### Proposed solution

I a saved draft is opened from `Choose from drafts`, it opens a form with pre-filled values.
- However if a user change his/her mind and doesn't want to continue with this draft and have a new empty form, there should be a way for the user to escape/cancel from editing the current draft. 

#### DoD (Definition of Done)

User is able to create a new form (from an existing form)

#### Testing

e2e test, peer to peer review

blankdots:
we can have a sticky bar that contains form controls such as:
1. clear form
2. new form
3. save draft
4. submit <object>

we don't need the `hide` button.

PR #148, opened 2021-02-02
Added skip-links to submission types when navigation with keyboard

### LINKED ISSUE #98: Fix keyboard tab order XML file upload, Choose file button
#### Describe the bug

This is continuation to Fix keyboard-based xml file upload #20. When the user has chosen to upload XML file, it would be nice to move focus directly to "Choose file" button instead of tabbing all the objectTypes in accordion first.

#### To Reproduce

1. Select an object type by tab and enter
2. Select submission type by tab, tab and enter
3. many tabs to reach "Choose file button"

#### Expected behavior

1. Select an objectType by tab and enter
2. Select submission type by tab tab and enter
3. Focus is on "Choose file"

#### A note
This could be checked with all the buttons.

saulipurhonen:
### Description

Use skip-links in object index when navigating with keyboard.
Pressing `enter` in `submission type` shows skip-link to submission. Pressing `enter` in skip-link focuses `hide` button, for now. This is going to change when new submission controls are introduced in #135 

### Related issues

Fixes #98

### Type of change

### Changes Made

- Slice for focus state
- Skip-links functionality in object index and object card components

### Testing

- [x] E2E tests

### DIFF
--- a/cypress/integration/app.spec.js
+++ b/cypress/integration/app.spec.js
@@ -19,9 +19,12 @@ describe("Basic e2e", function () {
     cy.get("textarea[name='description']").type("Test description")
     cy.get("button[type=button]").contains("Next").click()
 
-    // Fill a Study form and submit object
+    // Skip-link
     cy.get("div[role=button]").contains("Study").click()
-    cy.get("div[role=button]").contains("Fill Form").click()
+    cy.get("div[role=button]").contains("Fill Form").type("{enter}")
+    cy.get("div[role=button]").contains("Skip to form")
+
+    // Fill a Study form and submit object
     cy.get("input[name='descriptor.studyTitle']").type("Test title")
     cy.get("select[name='descriptor.studyType']").select("Metagenomics")
 
--- a/src/components/NewDraftWizard/WizardComponents/WizardAddObjectCard.js
+++ b/src/components/NewDraftWizard/WizardComponents/WizardAddObjectCard.js
@@ -1,5 +1,5 @@
 //@flow
-import React from "react"
+import React, { useRef, useEffect } from "react"
 
 import Button from "@material-ui/core/Button"
 import Card from "@material-ui/core/Card"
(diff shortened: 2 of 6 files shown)


saulipurhonen:
I'm not absolutely sure that focus to `hide` button is the best option. I tried to focus first form input but I think that would require lot of work to create logic for finding that element.

Feel free to suggest a better focus point.

blankdots:
I don't have any good suggestions :disappointed: but would link this to #135 so that we know this needs to be fixed as part of that issue.

hannyle:
It works fine for me, we can do the changes with new submission controls later.

One thing I have noticed is that sometimes when I press `enter` in `submissionType`, it goes straight to focus on `Hide` button, which means the `skip-link` is ignored. 
This could happen in any `submissionType` after I have pressed `enter` in `Hide` button in a previous `submissionType`

saulipurhonen:
@blankdots, linked to issue. 

@hannyle,  I added `resetFocus` state method to `Hide` buttons `onClick` handler. This fixes the issue described in:

> One thing I have noticed is that sometimes when I press `enter` in `submissionType`, it goes straight to focus on `Hide` button, which means the `skip-link` is ignored.
> This could happen in any `submissionType` after I have pressed `enter` in `Hide` button in a **previous** `submissionType`

blankdots (approved):

hannyle (approved):

lilachic (approved):
Might cause more errors, see https://github.com/CSCfi/metadata-submitter-frontend/issues/151 .

What about shift+tab, to return from form ?

saulipurhonen:
> Might cause more errors, see #151 .

Seems not to be related to #151, since this occurs in branches without skip-link functionality.

> What about shift+tab, to return from form ?

Do you mean that there should be a `skip-back-to` kind of a functionality if user wants to navigate back from `submission `to `object index`? In that case we'd need to inform user that there are `keyboard shortcuts` available

lilachic:
About #151 , it is with tabbing and entering that I managed to reproduce warning, it has show in other branches occasionally.

I find difficulties to navigate with keyboard, expected to return with shift+tab, no need for that now, skip-link to form is ok.

Issue #151, opened 2021-02-05
Errors in Console when Clicking Buttons


lilachic:
#### Describe the bug

When moving with tab and selecting with enter, and maybe also using mouse there appears a warning in console. This happens occasionally, actual steps to produce is difficult to provide. With the steps later the error states:
```
index.js:1 Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    in FormContent (at WizardFillObjectDetailsForm.js:358)
    in div (created by ForwardRef(Container))
    in ForwardRef(Container) (created by WithStyles(ForwardRef(Container)))
    in WithStyles(ForwardRef(Container)) (at WizardFillObjectDetailsForm.js:357)
    in WizardFillObjectDetailsForm (at WizardAddObjectCard.js:104)
console.<computed> @ index.js:1
overrideMethod @ react_devtools_backend.js:2430
printWarning @ react-dom.development.js:88
error @ react-dom.development.js:60
warnAboutUpdateOnUnmountedFiberInDEV @ react-dom.development.js:23161
scheduleUpdateOnFiber @ react-dom.development.js:21169
dispatchAction @ react-dom.development.js:15660
(anonymous) @ WizardFillObjectDetailsForm.js:130
setInterval (async)
handleStart @ WizardFillObjectDetailsForm.js:129
keyHandler @ WizardFillObjectDetailsForm.js:141`
```
#### To Reproduce

Steps to reproduce the behaviour:

1. Go to create folder
2. Click on study with mouse
3. contiue with tab to experiment select with enter
3. tab to Upload XML file
4. See error in console

#### Expected behavior

No warnings in console

lilachic:
This seem to be reproduciable, this time different code lines, see below. 
Steps to produce:

1. log in
2. Create submission
3. Tab to focus in back, tab to next, tab to study, 
4. Enter to select study
5. Tab to fill from
6. Enter to select fill form
7. Tab to activate link
8. Click link 

```
    in FormContent (at WizardFillObjectDetailsForm.js:358)
    in div (created by ForwardRef(Container))
    in ForwardRef(Container) (created by WithStyles(ForwardRef(Container)))
    in WithStyles(ForwardRef(Container)) (at WizardFillObjectDetailsForm.js:357)
    in WizardFillObjectDetailsForm (at WizardAddObjectCard.js:104)
[...]

PR #152, opened 2021-02-08
Bugfix/too many api calls

### LINKED ISSUE #150: Multiple API-calls on home page
#### Describe the bug

Loading `Draft` and `Published` submissions sends API call for each folder found for user. This is really performance heavy operation and should be queried with maximum 2 calls.

#### To Reproduce

1. Save multiple Draft folders
2. Publish multiple different folders
3. Go to /home
4. Check browser dev tools for `Network` tab

#### Expected behavior

Users folders should be queried with minimum API calls.

saulipurhonen:
### Description

Home page used to make an API call for each folderID in user data. In this fix we get the data with single API call.

Original problem also introduced problems with async calls and component unmounting which resulted in React error messages for cleaning up `useEffect` -methods. This same behavior caused errors in auto-save feature in form submissions and was prone to memory leaks.

### Related issues

Fixes #150 and #151. Depending on https://github.com/CSCfi/metadata-submitter/pull/193

### Type of change

### Changes Made

- Get folder and map folder data with single `GET` call to `/folders`
- Fixed auto-save timer reset problem in form submission.

### Testing

- [x] Tests do not apply

### DIFF
--- a/cypress/integration/home.spec.js
+++ b/cypress/integration/home.spec.js
@@ -126,7 +126,7 @@ describe("Home e2e", function () {
     cy.get("button[type=button]").contains("Publish").click()
     cy.get('button[aria-label="Publish folder contents and move to frontpage"]').contains("Publish").click()
 
-    cy.reload()
+    cy.get("div", { timeout: 10000 }).contains("Logged in as:")
     // Click "See all" button to navigate to all published folders list
     cy.get("div.MuiCardActions-root", { timeout: 30000 })
       .last()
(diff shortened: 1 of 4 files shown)


saulipurhonen:
E2E test for `home.spec.app` doesn't seem to pass since the backend doesn't have the fix for folder data yet.

hannyle (approved):
Looks good to me. Much better now that we have only 1 api call to get all folders.

blankdots (approved):

Issue #167, opened 2021-02-17
WizardShowSummary lacks of Object type "Dataset"


hannyle:
#### Describe the bug

**Dataset** belongs to Objects list,  but it doesn't exist in the **Summary**

#### To Reproduce

Steps to reproduce the behaviour:

1. Go to 'Create Submission'
2. Fill in Folder name and description
2. Click 'Next'
3. See "Dataset" at the end of the list of Objects on left side
4. Click 'Next'
5. In the Summary there is no "Dataset" object anymore

#### Expected behavior

**Dataset** and its submitted objects shows in the Summary

blankdots:
my bad closed it without looking

Issue #158, opened 2021-02-10
Having a constant file to contain enum values


hannyle:
#### Description
Currently there are many variables that are manually assigned their values in some files. 
For example, `submissionType` can have values as `"form"` or `"XML"` or `"Form"` or `"xml"`
The values are inconsistent and we are kind of hard-coding them. This can cause more bugs and hard to keep track in the future if the application gets bigger.

#### DoD (Definition of Done)
Having a file contained `enums` to define a set of constants

#### Testing
Peer to peer

blankdots:
there is a json schema validation in the backend which has the `enum` https://github.com/CSCfi/metadata-submitter/blob/develop/metadata_backend/helpers/schemas/folders.json#L44-L61 I haven't tried but not sure sending different cases than what is specified would work. are we sure it is possible ?

Or this is just for front-end to have a list of predefined values ?

hannyle:
> there is a json schema validation in the backend which has the `enum` https://github.com/CSCfi/metadata-submitter/blob/develop/metadata_backend/helpers/schemas/folders.json#L44-L61 I haven't tried but not sure sending different cases than what is specified would work. are we sure it is possible ?
> 
> Or this is just for front-end to have a list of predefined values ?

Yes, it is just for front-end to have `enums`, instead of hard-coding like now. For example:
- Instead of:
> switch (months) {
	case "Jan":
	// Jan
	case "Feb":
	// Feb
	case "March":
	// Mar	
}

- I would have:

> const months = {
	JAN: "january",
	FEB: "february",
	MAR: "march",
}

> switch (season) {
	case seasons.JAN:
	// Jan
	case seasons.FEB:
	// Feb
	case seasons.MAR:
        // Mar
}

PR #168, opened 2021-02-18
Feature/constant files

hannyle:
### Description

Having a constant folder with files containing predefined values for some variables
Add **Dataset** to the summary of folder (bug https://github.com/CSCfi/metadata-submitter-frontend/issues/167)

### Related issues

https://github.com/CSCfi/metadata-submitter-frontend/issues/158
https://github.com/CSCfi/metadata-submitter-frontend/issues/167

### Type of change

### Changes Made

- Create `constants` folder and add files `folder.js`, `object.js`, `wizardStatus.js`
- Create `enums` for predefined values to use in front-end side (e.g. `ObjectTypes`, `FolderSubmissionStatus`)
- Replace hard-coded variables with `enums`

### Testing

- [x] Unit Tests
- [x] Integration Tests


src/constants/object.js
```diff
@@ -0,0 +1,38 @@
+export const ObjectTypes = {
```
blankdots, line 1: if we are going to add more schemas we might need to extend this. Should we be dynamically or maintain once we add more schemas ?

blankdots (commented):
LGTM just want to know how we can maintain this to keep updated with some schemas and data types that might change in the backend ?

src/constants/object.js
blankdots: `/schemas` in the backend will give you the current schemas

blankdots (approved):

hannyle:
I managed to get `schemas` from backend and converted them to `ObjectTypes` to use in front-end. It looked fine in the beginning, but it actually created more work than I had expected because I could only have them in `components` while there are cases I need the `ObjectTypes` in `unit tests` as well. 

Also there are `ObjectsArray` and a few other cases depend on `ObjectTypes` so it's better to have it as a static object like now. 

In the future if the schemas get bigger I think we can refactor it.

blankdots:
ok ... then we can have them as is now, with the note we need to keep them in sync if we update them.

Issue #128, opened 2021-01-11
Form renders multiple times


saulipurhonen:
#### Describe the bug

When rendering form in `WizardFillObjectDetailsForm`, component is rendered multiple times on initialization.

#### To Reproduce

Steps to reproduce the behaviour:

1. Log anything in `FormContent` method.
2. Select `Fill Form` submission type 
3. Log does multiple times

#### Expected behavior

Form should be rendered only once for performance reasons.
It seems that multiple state changes in `WizardFillObjectDetailsForm` method cause this. 
https://stackoverflow.com/questions/53574614/multiple-calls-to-state-updater-from-usestate-in-component-causes-multiple-re-re

PR #178, opened 2021-03-03
Bugfix/form renders multiple times

hannyle:
### Description

- Fix bug https://github.com/CSCfi/metadata-submitter-frontend/issues/128 that the form renders multiple times when we open it
- Prevent the form from re-rendering every second when we start typing something into it

### Related issues

https://github.com/CSCfi/metadata-submitter-frontend/issues/128

### Type of change

### Changes Made

- Combine multiple internal states of component `WizardFillObjectDetailsForm` into one state object
- Fix the logics of counting-up `timer` and setting `DraftAutoSave` for the form

### Testing

- [x] Unit Tests
- [x] Integration Tests

### DIFF
--- a/src/components/NewDraftWizard/WizardForms/WizardFillObjectDetailsForm.js
+++ b/src/components/NewDraftWizard/WizardForms/WizardFillObjectDetailsForm.js
@@ -461,7 +461,7 @@ const WizardFillObjectDetailsForm = (): React$Element<typeof Container> => {
   const [responseInfo, setResponseInfo] = useState([])
 
   /*
-   * Fetch json schema from either local storage or API, set schema and dereferenced version to component state.
+   * Fetch json schema from either session storage or API, set schema and dereferenced version to component state.
    */
   useEffect(() => {
     const fetchSchema = async () => {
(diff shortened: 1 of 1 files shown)


src/components/NewDraftWizard/WizardForms/WizardFillObjectDetailsForm.js
```diff
@@ -432,20 +441,57 @@ const FormContent = ({ resolver, formSchema, onSubmit, objectType, folderId, cur
...
-  const [formSchema, setFormSchema] = useState({})
-  const [validationSchema, setValidationSchema] = useState({})
   const [submitting, setSubmitting] = useState(false)
-  const dispatch = useDispatch()
-  const { folderId } = useSelector(state => state.submissionFolder)
   const [responseInfo, setResponseInfo] = useState([])
-  const currentObject = useSelector(state => state.currentObject)
+
+  /*
+   * Fetch json schema from either local storage or API, set schema and dereferenced version to component state.
```
blankdots, line 464:
```suggestion
   * Fetch json schema from either session storage or API, set schema and dereferenced version to component state.
```
nitpicking as we have not been using local storage for quite some time we should make it clear it is session storage

blankdots (commented):
The changes look good but the branch seems to be a tad out of date.

~@hannyle @saulipurhonen  can you rebase this on `develop` i am worried that some changes to `flow-bin` package with regard to types might not be caught on this branch~

- never mind i think it was my local env that said flow-bin was an older version

blankdots (approved):

lilachic (approved):

PR #179, opened 2021-03-03
Bugfix/general bugfixes

blankdots:
### Description

Forgotten fixes for deployment to production. Also optimisations for building the UI.

### Type of change

### Changes Made

1. don't require `husky` in prod install, also updated the package
   - needed to add `pinst` for this to make it easier to disable in CI and Docker

2. fix codeql not running properly
3. update cypress - not a bugfix but might as well do it.
4. prettier should ignore build directory
5. move cypress and prettier to `eslintConfig` to allow production build to build smoothly without depending on cypress and prettier as they are not required for going to production
6. fix npm version to 7.6.0 in order to allow proper packages handling for newer versions.
7. add eslint as a dev dependency - not required but want to make it explicit
8. fixed immer vulnerability by updating `react-scripts` to `4.0.3`
9. e2e now test production build to reflect what is deployed in Docker container so that there aren't any inconsistencies 

### Testing

- [x] Tests do not apply


lilachic:
Why twice  in e2etests.yml
Screenshot 2021-03-03 at 17 39 01

blankdots:
@lilachic 
> Why twice in e2etests.yml

two different artifacts need 2 different stages, I could not figure out a good way to show 2 different artifacts separately in the same stage, but I am open to suggestions

package.json
```diff
@@ -32,13 +32,20 @@
     "format:check": "prettier --check \"**/*.+(json|yml|yaml|css|md)\"",
     "flow:check": "flow check",
     "concurrently": "concurrently",
-    "postinstall": "husky install"
+    "postinstall": "node -e \"process.env.NODE_ENV != 'production' && process.exit(1)\" || husky install"
   },
   "eslintConfig": {
+    "env": {
```
blankdots, line 38: not sure this is correct, but went by the the documentation

saulipurhonen: This looks good to me. Tried to find if there's a node script for this case but couldn't find any.

package.json
```diff
@@ -32,13 +32,20 @@
     "format:check": "prettier --check \"**/*.+(json|yml|yaml|css|md)\"",
     "flow:check": "flow check",
     "concurrently": "concurrently",
-    "postinstall": "husky install"
+    "postinstall": "node -e \"process.env.NODE_ENV != 'production' && process.exit(1)\" || husky install"
```
saulipurhonen, line 35: One option could be `prepublish` script but this seems to be ok too.

blankdots: changed :+1:

package.json
```diff
@@ -32,7 +32,7 @@
     "format:check": "prettier --check \"**/*.+(json|yml|yaml|css|md)\"",
     "flow:check": "flow check",
     "concurrently": "concurrently",
-    "postinstall": "node -e \"process.env.NODE_ENV != 'production' && process.exit(1)\" || husky install"
+    "prebublish": "husky install"
```
hannyle, line 35: There's a small typo "prepublish"

blankdots: good catch, wonder why it worked ...

blankdots: @saulipurhonen i will revert to `postinstall` `prepublish`  gives:

```
➜ npm install --production                             

> metadata-submitter-frontend@0.8.0 prepublish
> husky install

sh: husky: command not found
npm ERR! code 127
npm ERR! path /home/path/metadata-submitter-frontend
npm ERR! command failed
npm ERR! command sh -c husky install
```

saulipurhonen: Alright.
Just checked **Husky**  documentation and it's installed via `prepare`, not `prepublish`

blankdots: ok, so is it ok to leave like that, `prepare` gives the same or any other ideas ?

saulipurhonen: `prepare` seems to work locally via `npm install --production` but this I think this isn't that important issue. Only perk I see is is that it looks lot of cleaner.

hannyle: From the npm docs I just read, `prepublish` seems to be deprecated and `prepare` is another option.

blankdots: figured out another solution with `pinst` https://github.com/typicode/husky/issues/867#issuecomment-782845423 - lets see what CI has to say

blankdots: ... and it seems to like it. I will try it in docker the same, and it keeps it cleaner, and I will think where to document this in a future PR.

saulipurhonen (approved):
Looks good to me.

lilachic (approved):
Looks good to me. I don't know better way to give artifacts. Updating packages works.

Issue #174, opened 2021-03-01
Display titles and info instead of accession IDs


blankdots:
#### Description

Each of the objects currently displays the accessionID which makes it difficult for a user to identify the object submitted
![image](https://user-images.githubusercontent.com/47524/109462901-d98ec080-7a6c-11eb-955f-a7c2cc582116.png)

Similar for drafts:
![image](https://user-images.githubusercontent.com/47524/109463017-0511ab00-7a6d-11eb-9e4c-b4b8b12d493b.png)

Use the Title or an appropriate key: value pair from each object to display information about the submitted/draft object to the user.
If necessary display accessionID as secondary information.

#### DoD (Definition of Done)

Make the submitted/draft objects easily findable

#### Testing

Peer review and unit/integration tests.

PR #187, opened 2021-03-17
Feature/display object titles

hannyle:
### Description

- **submitted** and **draft** objects are displayed by `title` and `accessionId`, instead of showing only their `accessionId`

### Related issues

https://github.com/CSCfi/metadata-submitter-frontend/issues/174

### Type of change

### Changes Made

- Assign `form's title` or `xml's fileName` as value of `displayTitle` in `tags: {  displayTitle: " "}` when `create` object or `patch` folder
- Include `displayTitle` and `accessionId` for `ListItem` in components `WizardSavedObjectsList`, `WizardShowSummaryStep`, and `WizardDraftObjectPicker`
- Create a new file `utils/index.js` to contain reusable functions
- Fix logics for functions `saveDraft()` and `submit()` with `tags` 
- Update Unit tests for changes in related components

### Testing

- [x] Unit Tests
- [x] Integration Tests

### DIFF
--- a/src/components/NewDraftWizard/WizardComponents/WizardSavedObjectsList.js
+++ b/src/components/NewDraftWizard/WizardComponents/WizardSavedObjectsList.js
@@ -108,7 +108,7 @@ const WizardSavedObjectsList = ({ submissions }: WizardSavedObjectsListProps): R
                     className={classes.listItemText}
                     primary={getItemPrimaryText(item)}
                     secondary={item.accessionId}
-                    data-testid={item.schema}
+                    data-schema={item.schema}
                   />
                   <ListItemSecondaryAction>
                     <WizardSavedObjectActions
--- a/src/components/NewDraftWizard/WizardComponents/WizardSavedObjectsList.js
+++ b/src/components/NewDraftWizard/WizardComponents/WizardSavedObjectsList.js
@@ -108,6 +108,7 @@ const WizardSavedObjectsList = ({ submissions }: WizardSavedObjectsListProps): R
                     className={classes.listItemText}
                     primary={getItemPrimaryText(item)}
                     secondary={item.accessionId}
+                    data-testid={item.schema}
                   />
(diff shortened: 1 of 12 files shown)


blankdots (commented):
Looks good and I think it is what is needed.

Just one question I had that stopped from approving. the PR mentions `Integration Tests` but I don't see any added and should we add some ?

hannyle:
> Looks good and I think it is what is needed.
> 
> Just one question I had that stopped from approving. the PR mentions `Integration Tests` but I don't see any added and should we add some ?

Hm, I keep the `Integration Tests` there by mistake. But it's good that you mentioned, I was thinking about it and I will add some.

src/components/NewDraftWizard/WizardComponents/WizardSavedObjectsList.js
```diff
@@ -90,31 +93,37 @@ const WizardSavedObjectsList = ({ submissions }: WizardSavedObjectsListProps): R
...
+              className={classes.cardHeader}
+            />
+            <List aria-label={group.submissionType}>
+              {group.submittedItems.map(item => (
+                <ListItem key={item.accessionId} className={classes.objectListItem}>
+                  <ListItemText
+                    className={classes.listItemText}
+                    primary={getItemPrimaryText(item)}
+                    secondary={item.accessionId}
+                    data-testid={item.schema}
```
blankdots, line 111: would prefer if we have some like `data-schema` instead of `data-testid` as it makes more sense
the same goes for the other instances of this

blankdots: or `data-schemaType`

hannyle: I changed it to `data-schema`

blankdots (approved):

lilachic (approved):
Looks good to me.

Issue #91, opened 2020-11-09
Submission form status with browser buttons


saulipurhonen:
#### Proposed solution

App goes back to /home if user navigates with browser back button from any wizard step. To prevent this all wizard steps should be navigable with router (e.g. /newdraft/0 for folder creation).
Keep folder / submission status state if user navigates with browser back or forward buttons.

#### DoD (Definition of Done)

Browser back button navigates to previous step and app keeps state.

saulipurhonen:
If we want to keep URL as clean as possible, we could use local- or session storage for keeping the state. Check out https://github.com/rt2zz/redux-persist

PR #204, opened 2021-03-23
Feature/enable wizard navigating routes

lilachic:
### Description

Enable using of browser's back and forward buttons. Steps are navigable with router.

### Related issues

This solves the issue https://github.com/CSCfi/metadata-submitter-frontend/issues/91 .

### Type of change

- [ ] New feature (non-breaking change which adds functionality)

### Changes Made

Replaces using Redux global state step  with useParams hook, to pass relevant step of wizard. Steps are managed via routes.

### Testing

Testing is refactored to handle changed routes, both unit tests and integration tests with Cypress. There is no  additional new tests.

- [x] Unit Tests
- [x] Integration Tests
- [ ] Tests do not apply


src/App.js
```diff
@@ -66,7 +66,7 @@ const App = (): React$Element<typeof React.Fragment> => {
       <CssBaseline />
       <NavigationMenu />
       <Switch>
-        <Route path="/newdraft">
+        <Route path="/newdraft/:step">
```
blankdots, line 69: I think just having the step might not be enough. is it possible to make it relative to also the folder id ?

lilachic: So first folderID then step in the route maybe. But where is folderID used here in wizard creating submission? And how to handle non existing folderID, when the first two steps -1 and 0 are in question, keeping folderID -1? 
folderID is kept in global state, submissionFolderSlice. And in route /home/drafts or /home/published .
Maybe not in this PR?

blankdots: i assume the non-existing folderID would only be the case when `/newdraft` is first accessed and when a request is made to the folder API that is when we know it and we can use it in the URL (and included it with the previous and next links).

i was asking if possible, sure part of another PR is ok, I was thinking of what that would imply and if the changes in this PR would need to be redone. As such a feature might be necessary when we want to link folders from `/home/drafts` to be edited in the wizard. 

Does not have to be the only solution, however evaluating our options.

lilachic: One option with this task is to do it, is to keep steps in Redux state and keep route clean. So no steps in route and try to handle browsers history other wise.

saulipurhonen: Id's suggest on navigating with query parameters. Example URL with params:
`/newdraft?step=1&folderid=123123`

src/components/NewDraftWizard/WizardComponents/WizardStepper.js
```diff
@@ -149,21 +151,21 @@ const WizardStepper = ({ createFolderFormRef }: { createFolderFormRef?: CreateFo
         disableElevation
         color="primary"
         variant="outlined"
-        disabled={wizardStep < 1}
+        disabled={urlStep < 1}
```
saulipurhonen, line 154: I think `wizardStep` corresponds better to what this variable stands for.

src/App.js
saulipurhonen: Also, check and handler for `folderId` in query params would fix current problem where user refreshes the browser.

src/views/NewDraftWizard.js
```diff
@@ -50,7 +52,7 @@ const getStepContent = (wizardStep: number, createFolderFormRef: CreateFolderFor
     case 2:
       return <WizardShowSummaryStep />
     default:
-      throw new Error("Unknown step")
+      throw new Error("Unknown step") //THIS ERROR IS NOT CATCHED
```
blankdots, line 55: could this be solved as part of this PR, maybe 404 would make sense with the meaning: "a step that does not exist is being accessed."

src/components/NewDraftWizard/WizardSteps/WizardFrontpageStep.js
```diff
@@ -46,7 +44,8 @@ const NewObjectTooltip = withStyles(theme => ({
  */
 const WizardFrontpageStep = (): React$Element<typeof Grid> => {
   const classes = useStyles()
-  const dispatch = useDispatch()
+  //const dispatch = useDispatch()
```
blankdots, line 47: forgotten comment ?

blankdots (changes requested):
other than these small comments look good.

src/views/NewDraftWizard.js
blankdots: we can redirect to `Error404` page

lilachic: Maybe, I made a separate issue about this: https://github.com/CSCfi/metadata-submitter-frontend/issues/220 .

blankdots: i would like this part of this PR, `401` is for `Unauthorized` which does not make sense as the user is not trying to access something (s)he does not have permissions for but something that does not exist. So 404 makes more sense.

also don't leave comments in the code, if something is not finished put in issues, leave the code clean.

src/components/NewDraftWizard/WizardSteps/WizardFrontpageStep.js
saulipurhonen: Comment deleted

src/views/NewDraftWizard.js
saulipurhonen: Step now defaults to 404

blankdots (approved):

hannyle (approved):
Looks great to me!

PR #219, opened 2021-03-30
move ajv to v8+ dataPath -> instancePath

blankdots:
### Description

Switching to v8 introduced some issues: https://github.com/ajv-validator/ajv/blob/master/docs/v6-to-v8-migration.md#new-features

```
JSON Schema validation errors changes:

    dataPath property replaced with instancePath
    "should" replaced with "must" in the messages
    property name is removed from "propertyName" keyword error 
      message (it is still available in error.params.propertyName).

```
### Type of change

### Changes Made

`dataPath` property replaced with `instancePath`

### Testing

- [x] Tests do not apply

### Mentions

I am not sure about `propertyName` but does not seem there is any error, or we are not using the `name` and we are using the `error message` from `propertyName`.


blankdots:
closing as this does not seem to be the case, we do need an integration tests for this.

saulipurhonen (approved):
Tested and works great. Might need to clear browser cache

PR #222, opened 2021-03-31
Test blur on form field

### LINKED ISSUE #221: Test form field blur
#### Proposed solution

We don't test for blur events on form fields. These events can cause errors with validation and should be tested.

#### DoD (Definition of Done)

Blur events are tested.

saulipurhonen:
### Description

AJV package update revealed untested piece on form field validation. 
Validation triggers on blur and this event should be tested.

### Related issues

Closes #221 

### Type of change

### Changes Made

- Test blur event in `WizardFillObjectDetailsForm.test.js`
- Check that data exists in forms `handleChange` method
- Added test id to get the input from Material Textfield

### Testing

- [x] Integration Tests

### DIFF
--- a/src/__tests__/WizardFillObjectDetailsForm.test.js
+++ b/src/__tests__/WizardFillObjectDetailsForm.test.js
@@ -2,7 +2,7 @@ import React from "react"
 
 import "@testing-library/jest-dom/extend-expect"
 import { ThemeProvider } from "@material-ui/core/styles"
-import { render, screen, waitFor } from "@testing-library/react"
+import { fireEvent, render, screen, waitFor } from "@testing-library/react"
 import { Provider } from "react-redux"
 import configureStore from "redux-mock-store"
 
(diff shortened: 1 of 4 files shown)


hannyle (approved):
Looks good to me.

For **_Check that data exists in forms handleChange method_** I think we also have the function `checkFormCleanedValuesEmpty`. `clone` returns an object so I wonder if if can check the data exists.

saulipurhonen:
> Looks good to me.
> 
> For **_Check that data exists in forms handleChange method_** I think we also have the function `checkFormCleanedValuesEmpty`. `clone` returns an object so I wonder if if can check the data exists.

In this particular case input field rises error in the test case if `clone` is undefined. It seems that deep clone takes some time and the variable wasn't ready in the next assignment. Check for `clone` status gets past this error.

blankdots (approved):

lilachic (approved):

PR #233, opened 2021-04-07
DAC is an acronym treat it as one

blankdots:
### Description

We need to be able to process the object type as they are defined in the schemas yet still maintain them consistent through the UI. `DAC` is an acronym and should be handled as one.

### Related issues

My OCD kicked in regarding this and needed to solve it

### Type of change

### Changes Made

process the `objecttype` in `utils` so that we can better handle acronyms and have a consistent labels

### Testing

- [x] Tests do not apply

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -35,7 +35,7 @@ export const useQuery = (): URLSearchParams => {
 }
 
 export const formatDisplayObjectType = (objectType: string): string => {
-  if (objectType === "dac"){
+if (objectType === ObjectTypes.dac){
     return `${objectType.toUpperCase()}`
   } else {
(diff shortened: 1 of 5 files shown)


src/utils/index.js
```diff
@@ -33,3 +33,11 @@ export const getItemPrimaryText = (item: ObjectInsideFolderWithTags): string =>
 export const useQuery = (): URLSearchParams => {
   return new URLSearchParams(useLocation().search)
 }
+
+export const formatDisplayObjectType = (objectType: string): string => {
+  if (objectType === "dac"){
```
hannyle, line 38:
```suggestion
if (objectType === ObjectTypes.dac){
```

hannyle: just this small thing

hannyle (approved):

Issue #155, opened 2021-02-08
select drafts to keep as user templates before folder is published


blankdots:
#### Description

When a folder is published all drafts are deleted. A user might want to keep some as templates to reuse in other submissions.
Follow-up of https://github.com/CSCfi/metadata-submitter/issues/160

At the submission summary or in the drafts user would be able to select drafts to keep as templates

#### DoD (Definition of Done)

User can keep some templates as drafts and reuse them in submissions

#### Testing

unit and integration testing

blankdots:
User should be able to review drafts before selecting which one to save.

Issue #141, opened 2021-01-26
List draft template belonging to a user


blankdots:
#### Description

as pointed out https://github.com/CSCfi/metadata-submitter-frontend/issues/115#issuecomment-758067489 some of the drafts after a folder is published will be illustrated as templates we need to be able to distinguish them and reuse them in submissions

#### DoD (Definition of Done)

List on the home page with draft templates under `user` -> `drafts`
reuse them in submission process

#### Testing

integration and unit testing

blankdots:
should come after #155

PR #234, opened 2021-04-07
Feature/draft selections

hannyle:
### Description

- Drafts will be shown to users to save as templates before  publishing folders

### Related issues

Closes https://github.com/CSCfi/metadata-submitter-frontend/issues/155

### Type of change

### Changes Made

- Add a form for selecting drafts in component `WizardDraftSelections`
- Add option to `View` the draft
- Add `Cancel` button to close the `Publish dialog`
- Add e2e test for showing draft selections
- Fix e2e test for error 401

### Testing

- [x] Integration Tests

### Mentions

List of drafts saved as templates will be implemented in the next PR (for issuse https://github.com/CSCfi/metadata-submitter-frontend/issues/141)

### DIFF
--- a/src/components/NewDraftWizard/WizardComponents/WizardAlert.js
+++ b/src/components/NewDraftWizard/WizardComponents/WizardAlert.js
@@ -95,7 +95,6 @@ const CancelFormDialog = ({
   }
 
   let [dialogTitle, dialogContent] = ["", ""]
-  // let draftSelections = null
   let dialogActions
   const formContent = "If you save form as a draft, you can continue filling it later."
   const xmlContent = "If you save xml as a draft, you can upload it later."
--- a/src/components/NewDraftWizard/WizardComponents/WizardAlert.js
+++ b/src/components/NewDraftWizard/WizardComponents/WizardAlert.js
@@ -223,12 +223,7 @@ const CancelFormDialog = ({
           dialogTitle = "Publishing objects"
           dialogContent =
             "Objects in this folder will be published. Please choose the drafts you would like to save, unsaved drafts will be removed from this folder."
-
-          dialogActions = (
-            <DialogActions style={{ justifyContent: "center" }}>
-              <WizardDraftSelections onHandleDialog={handleDialog} />
-            </DialogActions>
-          )
+          dialogActions = <WizardDraftSelections onHandleDialog={handleDialog} />
           break
         }
         default: {
@@ -289,7 +284,6 @@ const CancelFormDialog = ({
       <DialogTitle id="alert-dialog-title">{dialogTitle}</DialogTitle>
       <DialogContent>
         <DialogContentText id="alert-dialog-description">{dialogContent}</DialogContentText>
-        {/* {draftSelections} */}
       </DialogContent>
       {error && <ErrorMessage message={errorMessage} />}
       {dialogActions}
--- a/src/components/NewDraftWizard/WizardComponents/WizardAlert.js
+++ b/src/components/NewDraftWizard/WizardComponents/WizardAlert.js
@@ -226,7 +226,7 @@ const CancelFormDialog = ({
 
           dialogActions = (
             <DialogActions style={{ justifyContent: "center" }}>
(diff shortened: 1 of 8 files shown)


src/services/errorMonitor.js
```diff
@@ -3,7 +3,7 @@ export const errorMonitor = res => {
   if (!res.ok) {
     switch (res.status) {
       case 401:
-        window.location = "/error401"
+        // window.location = "/error401"
```
blankdots, line 6: forgotten comment ?

src/components/NewDraftWizard/WizardComponents/WizardDraftSelections.js
```diff
@@ -0,0 +1,197 @@
...
+import { setSubmissionType } from "features/wizardSubmissionTypeSlice"
+import draftAPIService from "services/draftAPI"
+import type { ObjectInsideFolderWithTags } from "types"
+import { getItemPrimaryText } from "utils"
+
+const useStyles = makeStyles(theme => ({
+  formComponent: {
+    margin: theme.spacing(2),
+    padding: 0,
+    overflow: "scroll",
```
blankdots, line 29: image
There is some weird formatting I think we can change this to `auto` and it will be solved, but I have not tested with more items
```suggestion
    overflow: "auto",
```

blankdots (approved):
small issue other than that looks good, draft is saved under user :+1:

hannyle:
> small issue other than that looks good, draft is saved under user 👍

Looks like it's the responsive css issue since I don't have it on my machine.
You could check if it's fixed with my new commit.

blankdots (approved):
now it is better

lilachic:
It runs nicely through all the tests. But when selecting study, fill form, give title and select type and save that as draft, then selecting sample and xml and submit, click next and click publish. It'll open sample form to edit instead of study form? And clicking update for empty form gives patch error.

blankdots:
> update for empty form gives patch error.

Sounds like a bug for updating empty form, which i don't think it is related to this issue, but can be filled in as a bug.

blankdots:
although we do have tests for submitting empty form i assume we don't check updating empty form, maybe that bug can be solved as part of this issue :thinking: . any opinions ?

lilachic:
So now is that bug is filed as 245, or is that another?
What is there still left is that when I select a study from, enter title and select type, then select some other object, like sample, select XML file and submit that file, click next and click publish, then opens the select window, then check the study from to edit, click view, it opens sample form instead of study form?

blankdots:
@lilachic that #245 has nothing to do with this issue (i was working on `develop` when i discovered it), if you want to file a bug you can do it or you can talk with @hannyle who opened this PR to understand how it should be solved what you discovered.

hannyle:
> So now is that bug is filed as 245, or is that another?
> What is there still left is that when I select a study from, enter title and select type, then select some other object, like sample, select XML file and submit that file, click next and click publish, then opens the select window, then check the study from to edit, click view, it opens sample form instead of study form?

@lilachic I understand what you have mentioned. There are 2 issues here: 
- `View` draft redirects to a wrong `objectType`
- Empty form issue https://github.com/CSCfi/metadata-submitter-frontend/issues/245

I have pushed some fixes, please check again if if works correctly now.

src/components/NewDraftWizard/WizardComponents/WizardAlert.js
```diff
@@ -92,6 +95,7 @@ const CancelFormDialog = ({
   }
 
   let [dialogTitle, dialogContent] = ["", ""]
+  // let draftSelections = null
```
lilachic, line 98: Is the comment needed?

lilachic (approved):
Now it looks good.  Seems to work correctly.

hannyle:
> although we do have tests for submitting empty form i assume we don't check updating empty form, maybe that bug can be solved as part of this issue 🤔 . any opinions ?

I can add the test for `Update` empty form as part of issue `245`

PR #281, opened 2021-04-28
Feature/react-hook-form v7 migration

### LINKED ISSUE #260: React Hook Form version bump breaking changes
#### Proposed solution

React Hook Form version bump from 6 to 7 will cause breaking changes.
https://react-hook-form.com/migrate-v6-to-v7/

#### DoD (Definition of Done)

Implement steps explained in migration guide and upgrade React Hook Form module to version 7.

#### Testing

All tests pass with new version.

saulipurhonen:
### Description

React Hook Form migration from V6 to V7 introduced some breaking changes detailed in:
https://react-hook-form.com/migrate-v6-to-v7/

### Related issues

Should fix Dependabot PR #266. Closes #260 

### Type of change

### Changes Made

 - Update React Hook Form to V7
 - Use `Controller` with Material UI fields in `WizardCreateFolderStep`
 - Register file upload input as described in documentation
 - Render form fields as uncontrolled fields with missing refs
 - Omit patch values so that blocked fields are left out from the request.

### Testing

- Current tests should cover the PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,7 @@
         "lodash": "^4.17.21",
         "react": "^17.0.2",
         "react-dom": "^17.0.2",
-        "react-hook-form": "^6.15.5",
+        "react-hook-form": "^7.3.4",
         "react-redux": "^7.2.4",
         "react-router-dom": "^5.2.0",
         "react-scripts": "^4.0.3"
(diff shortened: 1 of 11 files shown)


saulipurhonen:
@blankdots, I encountered a problem with `PATCH` operations when updating saved and draft objects:
When patching, I'd get a response "Some items (e.g: accessionId, publishDate, dateCreated) cannot be changed.".
I fixed this with omitting the blocked values, but what I'm wondering is that why does this happen on this branch, but not on develop?

Any ideas?

blankdots:
@saulipurhonen 
> When patching, I'd get a response "Some items (e.g: accessionId, publishDate, dateCreated) cannot be changed.".

That is expected as these items cannot be updated to avoid any API call overwriting them, as these are relevant for the whole submission.

Why this happens on this branch and not on `develop` that is good question, not sure, even more so cannot tell, given that it is expected to be like that, and the changes this branch introduces has nothing to do with it. 

it is indeed a puzzling :thinking: - when I have time will try to investigate

lilachic (approved):

saulipurhonen:
> @saulipurhonen
> 
> > When patching, I'd get a response "Some items (e.g: accessionId, publishDate, dateCreated) cannot be changed.".
> 
> That is expected as these items cannot be updated to avoid any API call overwriting them, as these are relevant for the whole submission.
> 
> Why this happens on this branch and not on `develop` that is good question, not sure, even more so cannot tell, given that it is expected to be like that, and the changes this branch introduces has nothing to do with it.
> 
> it is indeed a puzzling 🤔 - when I have time will try to investigate

This seems to be related how React Hook Form functions with the `reset` and `getValues` methods in V7.
In V6 if data was set to form via `reset(currentObject)` method, it would only add properties that match to form fields. In V7 it adds all object fields to the form values.

hannyle:
Oops, I was almost done finishing this PR. It was merged already.
Anyway, there were quite many things to learn here from my side. As @saulipurhonen  mentioned above that RHF v7 has a lot of hidden changes. Methods `reset` and `getValues` are one of the examples. And also for `useFieldArray` `append` in `WizardJSONSchemaParser`, now the `append()` method will automatically add the default `field` for us, no need for `append({ items })` as v6.

blankdots:
@hannyle we can open a new PR if anything was missed, or track it via an issue if we will handle it later.

saulipurhonen:
> Oops, I was almost done finishing this PR. It was merged already.
> Anyway, there were quite many things to learn here from my side. As @saulipurhonen mentioned above that RHF v7 has a lot of hidden changes. Methods `reset` and `getValues` are one of the examples. And also for `useFieldArray` `append` in `WizardJSONSchemaParser`, now the `append()` method will automatically add the default `field` for us, no need for `append({ items })` as v6.

Oh, sorry! My mistake and should have double checked that there are more than one approve.

hannyle:
no worries :)

PR #296, opened 2021-05-06
Higlight required fields

### LINKED ISSUE #291: Highlight required fields
#### Describe the bug

Highlighting of required fields doesn't work properly.
It seems that `WizardJSONSchemaParser` doesn't handle `required` attribute as it should.

#### To Reproduce

1. Create new submission
2. Click Analysis
3.  `Analysis Type` dropdown isn't highlighted

#### Expected behavior

All required fields should be highlighted.

saulipurhonen:
### Description

All required fields are not highlighted and this can confuse user.

### Related issues

Closes #291 

### Type of change

### Changes Made

- Handle sections that are required but don't contain any required fields
- Highlight required text areas
- Highlight required One Of fields
- Asterisk after `FieldArray` heading if section is required

### Testing

- [x] Tests do not apply

### DIFF
--- a/cypress/integration/app.spec.js
+++ b/cypress/integration/app.spec.js
@@ -90,11 +90,12 @@ describe("Basic e2e", function () {
 
       // Analysis type
       cy.get("select[name='analysisType']").select("Reference Alignment")
-      cy.get("select[name='analysisType']", { timeout: 10000 }).should("have.value", "Reference Alignment")
-      cy.get("select[name='analysisType.assembly']").select("Standard")
-      cy.get("input[name='analysisType.assembly.accessionId']").type("Standard Accession Id")
-      cy.get("h3").contains("Sequence").parent().children("button").click()
-      cy.get("input[name='analysisType.sequence[0].accessionId']").type("Sequence Standard Accession Id")
+      cy.get("select[name='analysisType.referenceAlignment.assembly']").select("Standard")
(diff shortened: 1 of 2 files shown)


saulipurhonen:
Found some issues where non-required field were handled as required.
There's also some fields that need to be required but nesting loses the required attribute.

saulipurhonen:
Logic needs some fine tuning since highlighting of first field of deeply nested section (which ancestor is a required property) doesn't work.

In `example` schema, `Design` is required property and therefore it needs data to pass validation. 

Example in deeply nested field that should be required:
image

In above image user has selected `Pool` as `Sample Reference`. In this case first field of child section should be required.

If user chooses `Invidual Sample` from options, highlighting works since first field gets required property.


Also, duplicate heading for `Pool` doesn't look right in design perspective. It works as should if checked against schema.

hannyle:
@saulipurhonen I kinda fixed the issue with requiring and highlighting the first field of deeply nested object.
There are still some kinds of error with `oneOf` field (I guess) especially with `Analysis` object that even though I filled everything the form couldn't submit. We can investigate this more.

blankdots (approved):

hannyle (approved):
So the issue was actually `traverseFields` `object` argument in `FormOneOfFied`. 
It's good that you noticed this.

lilachic (approved):

Issue #172, opened 2021-03-01
Tooltip for Form fields


blankdots:
#### Description

As a user I would like to have more information on the form fields and understand how to fill them in.
The backend JSON schemas contain some descriptions for the fields and a tooltip on the fields would be helpful.

Some tooltips might need wrapping as they are too verbose.

#### DoD (Definition of Done)

Display description of the fields.

#### Testing

Unit test, maybe Integration testing if possible and Peer review.

blankdots:
will require some optimisation for rendering multiple descriptions

blankdots:
be aware that descriptions are also associated with some of the nesting elements of the fields, sections and subsections and that is relevant for understanding the fields.

PR #310, opened 2021-05-17
Feature/form field tooltips

lilachic:
### Description

A new feature to add description of fields to show as tooltips. Implemented by transferring the description of the field to title of the field. This is done by following the existing parsing of the forms fields from ENA schema.

### Related issues

Fixes https://github.com/CSCfi/metadata-submitter-frontend/issues/172

### Type of change

### Changes Made

Added description to the title of fields.

### Testing
This will need testing that if description exist it is rendered when hovering?

- [x] Unit Tests
- [ ] Integration Tests
- [ ] Tests do not apply

### Mentions
Now this fails in a unit test WizardFillObjectDetailsForm.test.js with `Unable to find an element by: [data-testid="descriptor.studyTitle"]`

### DIFF
--- a/src/components/NewDraftWizard/WizardForms/WizardJSONSchemaParser.js
+++ b/src/components/NewDraftWizard/WizardForms/WizardJSONSchemaParser.js
@@ -286,27 +286,27 @@ const FormOneOfField = ({ path, object, nestedField }: { path: string[], object:
         const error = _.get(errors, name)
         return (
           <div>
-              <TextField
-                name={name}
+            <TextField
+              name={name}
               inputProps={{"title": description}}
-                label={label}
-                defaultValue={field}
-                select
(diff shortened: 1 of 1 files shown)


src/components/NewDraftWizard/WizardForms/WizardJSONSchemaParser.js
```diff
@@ -394,18 +401,19 @@ const FormSelectField = ({ name, label, required, options, nestedField }: FormSe
   </ConnectForm>
 )
 
-/*
+/* !!! Analysis Sequence Variantion imputation checkbox does not have description
```
blankdots, line 404: yes that is true, if they ever add on in the ENA schemas we can add one as well, but please meanwhile these kind of comments either we create issues in backend or we keep them somewhere else, not in the code.

Issue #295, opened 2021-05-06
Form checkbox reset


saulipurhonen:
#### Describe the bug

Checkbox selections are not cleared when user clears the form with `Clear form` button.

#### To Reproduce

Steps to reproduce the behaviour:

1. Go to `Dataset` object type
2. Select options from checkbox array
3. Clear form with `Clear form` button
4. Selections persist

#### Expected behavior

Form should clear

saulipurhonen:
This problem is related to if user chooses to edit draft / submitted object. Checkboxes aren't selected although selections exist in object data.

PR #323, opened 2021-05-28
Bugfix/edit does not work for all fields

### LINKED ISSUE #256: Edit does not work for all fields
#### Describe the bug

Edit does not seem to work for all fields for `Sample`, `Experiment`, `Analysis` and `Run`

Please add integration tests to prevent this from happening

#### To Reproduce

1. Go to `Create Submission` to create folder
2. Click on an object `Sample` 
3. Fill in form with required filed and also `Sample Data Type`
4. Submit form
5. clear form
6. edit previous form and see how `Sample Data Type` does not get proper data
7. See screenshots for other schema types

#### Expected behavior

 `Sample Data Type` for `Sample` does gets proper data when edit is pressed
All fields get the right information to edit.

#### Screenshots

`Sample`

Peek 2021-04-19 13-15

----

`Experiment`

Peek 2021-04-19 14-28

----

`Run`

Peek 2021-04-19 14-30

----

`Analysis`
Peek 2021-04-19 14-31

saulipurhonen:
### Description

Some fields didn't populate correctly when editing saved / submitted form.

### Related issues

Closes #256 and #295 

### Type of change

### Changes Made

- Set default value for OneOf -field if there's a selection in object data
- Set checked status for Checkbox component
- Dedicated E2E tests for form editing

### Testing

- [X] E2E Tests

### DIFF
--- a/cypress/integration/app.spec.js
+++ b/cypress/integration/app.spec.js
@@ -1,15 +1,11 @@
 describe("Basic e2e", function () {
-  const baseUrl = "http://localhost:" + Cypress.env("port") + "/"
-
   it("should navigate to home with click of login button", () => {
-    cy.visit(baseUrl)
-    cy.get('[alt="CSC Login"]').click()
+    cy.login()
   })
 
   it("should create new folder, add Study form, upload Study XML file, add Analysis form, and publish folder", () => {
-    cy.visit(baseUrl)
-    cy.get('[alt="CSC Login"]').click()
-    cy.wait(1000)
+    cy.login()
+
     cy.get("button", { timeout: 10000 }).contains("Create Submission").click()
 
     // Navigate to folder creation
@@ -35,20 +31,6 @@ describe("Basic e2e", function () {
     // Submit form
     cy.get("button[type=submit]").contains("Submit").click()
     cy.get(".MuiListItem-container", { timeout: 10000 }).should("have.length", 1)
-
-    // Clear form
-    cy.get("button[type=button]").contains("Clear").click()
-
-    // Edit saved submission
-    cy.get("button[type=button]").contains("Edit").click()
-    cy.get("input[name='descriptor.studyTitle']").should("have.value", "New title")
-    cy.get("input[name='descriptor.studyTitle']", { timeout: 10000 }).focus().type(" edited").blur()
-    cy.get("input[name='descriptor.studyTitle']").should("have.value", "New title edited")
-    cy.get("button[type=button]").contains("Update").click()
-    cy.get("div[role=alert]").contains("Object updated")
-    cy.get("button[type=button]").contains("New form").click()
-    cy.get("button[type=button]").contains("Edit").click()
-    cy.get("input[name='descriptor.studyTitle']").should("have.value", "New title edited")
 
     // Upload a Study xml file.
     cy.get("div[role=button]").contains("Upload XML File").click()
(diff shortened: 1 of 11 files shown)


hannyle (commented):
I can see that the issue with `Sample` form has been fixed.
However for the other forms: `Analysis` and `Run`, the forms still miss some nested fields when editing.

saulipurhonen:
> I can see that the issue with `Sample` form has been fixed.
> However for the other forms: `Analysis` and `Run`, the forms still miss some nested fields when editing.
> 
> For example: in `Analysis`:
> 
> 1. Select `Reference Alignment` for `Analysis Type`
> 2. Select `Standard` for `Reference assembly details`
> 3. Fill in the field `Accession Id` of `Standard`
> 4. Click `Save as Draft`
> 5. Edit this draft
> 6. The fields append to `Standard` is missing

Good catch. For some reason I missed this one and the one in the `Run`. `Sample` and `Experiment` should work but I need to look into why it doesn't append fields in all cases.

saulipurhonen:
All other fields except deeply nested `fieldArray` fields should work.

`Experiment` form as an example:

1. Add new item to `Expected Base Call Table` and fill elevated field array.
2. Save draft and click edit from `Choose from drafts`.

Added item in `Expected Base Call Table` isn't rendered but data is there if you look into `currentObject` in state.

[In React Hook Form documentation for `useFieldArray`](https://react-hook-form.com/api/usefieldarray) it seems that to get data from right registered field, we need to use whole path instead of last path element.

Path to element in this case would be: `design.spotDescriptor.readSpec.expectedBaseCallTable`
Instead we're trying to use only `expectedBaseCallTable` for element lookup and this doesn't match.

#322 fixed issue when user tries to edit form if deeply nested values exist. This might however cause issues when we'd like to render those deeply nested values. Fix passes only the last item of path to `useFieldArray` and I'm thinking this could be the culprit for this issue.

saulipurhonen:
Found another problem which is:
Experiment > Expected Base Call Table > Processing > Complex Processing > Pipeline > Pipe Section

All other fields populate values correctly except `Prev Step Index`.

saulipurhonen:
> Found another problem which is:
> Experiment > Expected Base Call Table > Processing > Complex Processing > Pipeline > Pipe Section
> 
> All other fields populate values correctly except `Prev Step Index`.

I'm not 100% sure but the data from saved submission looks odd to me:
```
{
  "processing": [
    {
      "pipeline": {
        "pipeSection": [
          {
            "prevStepIndex": "test value"
          }
        ]
      }
    }
  ]
}
```

And in schema `prevStepIndex` is an oneOf typed property and has two options: `String value` and `Null value`

But it seems there's no way to tell to which option should get the value of `prevStepIndex`.

@blankdots, any ideas on this? Or am I missing something?

blankdots (commented):
all looks good to me, only thing i noticed is this error otherwise i would have approved.

Peek 2021-06-22 16-31

to replicate
1. Select Single Processing first
2. Select Complex Processing
3. see error

saulipurhonen:
> all looks good to me, only thing i noticed is this error otherwise i would have approved.
> 
> Peek 2021-06-22 16-31
> 
> to replicate
> 
> 1. Select Single Processing first
> 2. Select Complex Processing
> 3. see error

Found this problem yesterday and created a new issue for it: #351 
I think we need to unregister and re-register oneOf field when changing options.

blankdots:
@saulipurhonen can #351 be done as part of this issue, or we approve this one as is and continue with another PR for that
either is fine for me.

saulipurhonen:
> @saulipurhonen can #351 be done as part of this issue, or we approve this one as is and continue with another PR for that
> either is fine for me.

I think this behavior is somewhat related to the original issue, but I think we should handle it as a separate PR to keep review easier for all.

blankdots (approved):

hannyle (approved):
Looks good to me :)

Issue #58, opened 2020-08-10
figure out linking of accession IDs


blankdots:
#### Description

As accession IDs are generated in the backend do not display them on the frontend, however some of the objects related to one another and they should be linked via their accession IDs

![image](https://user-images.githubusercontent.com/47524/109460175-dbef1b80-7a68-11eb-94e3-a1510872215e.png)

Note that a `Study` can have multiple `Dataset`s

more info on metadata objects:
- https://ena-docs.readthedocs.io/en/latest/submit/general-guide/metadata.html

#### DoD (Definition of Done)

Have a drop down to select from existing `acessionID`s from a folder, but with consideration of the linking.
Link elements by their `acessionID`s so that it reflects the ENA schema we are using

#### Testing

peer review and maybe unit testing

blankdots:
if we establish a link in the front-end we put as a tag in the backend element example: `tags` -> `links` -> array which has:
```
{"acccessionId": "87693408", "schema": "study"}
```

blankdots:
actually tags might not be needed, but we can just use the corresponding fields

PR #330, opened 2021-06-02
Feature/linking accession ids

hannyle:
### Description

- `accessionIds` of related objects are now linked. 
- Instead of typing manually, inside each form we can select `accessionIds` of other submitted objects according to ENA Metadata model: https://metadata-submitter.readthedocs.io/en/docs-metadata-and-auth/metadata.html

### Related issues

https://github.com/CSCfi/metadata-submitter-frontend/issues/58

### Type of change

### Changes Made

- Modify `dereferencedSchema` in `WizardFillObjectDetailsForm` to add arrays of `accessionId`  of other `ObjectTypes`
- Add e2e test for testing linking `accessionIds` of all `ObjectTypes` 
- Update unit test

### Testing

- [x] Unit Tests
- [x] Integration Tests

### Mentions

- In Analysis form, there is a section for `Analysis Reference` and inside it there is a field for Accession Id. This could mean that Analysis could link to each other (?)
- Users may confuse when they see only the `accessionIds` without knowing exactly which submission is which. We can add the `title` there for ones that have `title`.


saulipurhonen (approved):
Is it on purpose with `FieldArray` fields to automatically choose reference accession ID when clicking `Add new item`?
For example in `Run` > `Experiment Reference`

If selection field is visible on initial render, default value is empty. Eg. `Experiment` > `Study Reference`

I think it's ok to work this way and if we want to change it, we can do it afterwards.

Otherwise LGTM!

hannyle:
> `FieldArray` fields to automatically choose reference accession ID when clicking `Add new item` 

--> This is not on purpose. I think it shouldn't render any selection in the beginning. It can be fixed in another PR.

blankdots:
Looks good! however the accession id might be something that users might find difficult to identify.
Is there a way to also put in the drop down the title near the id - but when it is selected just put in the field the accession id.
something like:

drop down :arrow_down_small: 
 359625734 (study 1)
 589675476 (study 2)
 
When user selects `589675476 (study 2)` in the field only `589675476` remains 

This is not mandatory and could be part of another PR. What do you think?

cypress/integration/app.spec.js
```diff
@@ -107,7 +107,7 @@ describe("Basic e2e", function () {
       cy.get("input[name='experimentRef[0].identifiers.submitterId.value']").type("Experiment Test Value")
 
       // Study
-      cy.get("input[name='studyRef.accessionId']").type("Study Test Accession Id")
+      // cy.get("input[name='studyRef.accessionId']").type("Study Test Accession Id")
```
blankdots, line 110: forgotten comment ?

blankdots (approved):

norling:
Hi! 

I did some limited testing, and had some trouble.
To replicate, create a new submission:
- create and submit a sample
- click "Experiment"->"Fill Form"

The site crashes with:
```
Uncaught (in promise) TypeError: s.filter is not a function         WizardJSONSchemaParser.js:356:39
    children WizardJSONSchemaParser.js:356
    _r WizardJSONSchemaParser.js:128
    React 7
    unstable_runWithPriority scheduler.production.min.js:18
    React 5
    e WizardFillObjectDetailsForm.js:466
    l runtime.js:63
    _invoke runtime.js:293
    S runtime.js:118
    Babel 2
```
Tested in firefox 89 and chrome 91.0.4472.101.

hannyle:
> Looks good! however the accession id might be something that users might find difficult to identify.
> Is there a way to also put in the drop down the title near the id - but when it is selected just put in the field the accession id.
> something like:
> 
> drop down 🔽
> 359625734 (study 1)
> 589675476 (study 2)
> 
> When user selects `589675476 (study 2)` in the field only `589675476` remains
> 
> This is not mandatory and could be part of another PR. What do you think?

This is a good idea. We can put the `title` next to `AccessionId`.
I also think that in some of submitted forms, they may not have the `title` because the `title` is not required, then we can only show the `AccessionId`. Or with the case of `DAC form`, we then need to show the title as `Main Contact: Name`.

hannyle:
> Hi!
> 
> I did some limited testing, and had some trouble.
> To replicate, create a new submission:
> 
> * create and submit a sample
> * click "Experiment"->"Fill Form"
> 
> The site crashes with:
> 
> ```
> Uncaught (in promise) TypeError: s.filter is not a function         WizardJSONSchemaParser.js:356:39
>     children WizardJSONSchemaParser.js:356
>     _r WizardJSONSchemaParser.js:128
>     React 7
>     unstable_runWithPriority scheduler.production.min.js:18
>     React 5
>     e WizardFillObjectDetailsForm.js:466
>     l runtime.js:63
>     _invoke runtime.js:293
>     S runtime.js:118
>     Babel 2
> ```
> 
> Tested in firefox 89 and chrome 91.0.4472.101.

Hi! I hope my latest commit would fix this issue.

cypress/integration/linkingAccessionIds.spec.js
```diff
@@ -0,0 +1,266 @@
...
+      .siblings()
+      .within(() =>
+        cy
+          .get("div[role=button]")
+          .contains("Fill Form", { timeout: 10000 })
+          .should("be.visible")
+          .then($btn => $btn.click())
+      )
+    cy.get("input[name='sampleName.taxonId']").type(123)
+    // Submit Study form
```
lilachic, line 48: or Submit Sample form?

hannyle: yep, my mistake. I fixed the comments.

blankdots (approved):

PR #356, opened 2021-06-24
Feature/form field tooltips

### LINKED ISSUE #172: Tooltip for Form fields

lilachic:
### Description

A new feature to add description of fields to show as tooltips. Implemented by transferring the description of the field to the icon title tooltip by the field. This is done by following the existing parsing of the forms fields from ENA schema.

### Related issues

Fixes #172

### Type of change

### Changes Made
Added Material UI tooltip and Material-UI help icon at /WizardJSONSchemaParser . 
Added check for tootipWizardFillObjectDetailsForm.test.js

### Testing

- [x] Unit Tests
- [ ] Integration Tests
- [ ] Tests do not apply

### DIFF
--- a/cypress/integration/editForm.spec.js
+++ b/cypress/integration/editForm.spec.js
@@ -90,7 +90,7 @@ describe("Populate form and render form elements by object data", function () {
 
     // Switch to select and fill Complex processing
     cy.get("select[name='processing']").select(testData.complexProcessing)
-    cy.get(":nth-child(7) > .array > .MuiButtonBase-root > .MuiButton-label").click()
+    cy.get("h2[data-testid='processing']").parent().children("button").click()
     cy.get(".MuiPaper-root > :nth-child(1) > .formSection > .array > .MuiButtonBase-root > .MuiButton-label").click()
     cy.get("input[data-testid='processing[0].pipeline.pipeSection[0].stepIndex']").type(testData.stepIndex)
     cy.get("select[name='processing[0].pipeline.pipeSection[0].prevStepIndex']").select(testData.stringValue)
@@ -137,7 +137,7 @@ describe("Populate form and render form elements by object data", function () {
     cy.get("button[aria-label='Continue draft']").first().click()
 
     // Test that Prev Step Index is nulled
-    cy.get("input[data-testid='processing[0].pipeline.pipeSection[0].stepIndex']").should("not.exist")
+    cy.get("input[data-testid='processing[0].pipeline.pipeSection[0].prevStepIndex']").should("not.exist")
   })
 
   it("should render Run form correctly when editing", () => {
--- a/cypress/integration/linkingAccessionIds.spec.js
+++ b/cypress/integration/linkingAccessionIds.spec.js
@@ -98,8 +98,7 @@ describe("Linking Accession Ids", function () {
 
     // Run form
     cy.get("div[role=button]").contains("Run").click()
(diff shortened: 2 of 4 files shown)


blankdots (commented):
Before fully approving: 

is it possible to align vertically the question mark with the form field so that they look on the same line ? right now they look a bit out of place at least in my version of firefox
image
Also for DAC (and maybe other cases) the positioning is also odd, do you think we can fix it ?
image

Other suggestions of improvement are:
- maybe we can place the tool tip oriented on the top - https://material-ui.com/components/tooltips/#positioned-tooltips (top-end could work)
- in another issue maybe we can add tool tips for different sections of the form as well for example `Experiment > Study Reference` the `Study Reference` has a description maybe we use that (again this can be a different issue)
- i have low luminosity on my screen and the grey of the question mark is very faded, maybe we can give it a darker grey e.g. `rgba(0, 0, 0, 0.40)` or something similar

blankdots:
some cypress e2e tests would be nice, not required but nice.

lilachic:
>     * i have low luminosity on my screen and the grey of the question mark is very faded, maybe we can give it a darker grey e.g. `rgba(0, 0, 0, 0.40)` or something similar

I wanted the question mark be less disturbing, when a user knows what to do, by choosing the same colour than in the framing of the fields, it depends what is wanted?

blankdots:
> I wanted the question mark be less disturbing, when a user knows what to do, by choosing the same colour than in the framing of the fields, it depends what is wanted?

i think we need to guide here by the accessibility principles  https://webaim.org/articles/contrast/  i see it recommends: `Gray (#767676) on white`
we can use: https://wave.webaim.org/extension/ to test this and get suggestions - which apparently only notices that we have that hidden

blankdots:
oh ... we need to use a different tool: https://webaim.org/resources/contrastchecker/

image

lilachic:
> is it possible to align vertically the question mark with the form field so that they look on the same line ? right now they look a bit out of place at least in my version of firefox
Easy one done in latest commit `replace help icon inline with textfield `. More difficult are the ones inside some box, did not found those nor the solution yet.

blankdots (approved):
I approve as is. Suggestions of improvement are left at the decision of the PR creator to continue in this PR or following with other PRs.

hannyle (approved):
Looks good to me. 
CSC Design System also has some suggestions for `Tooltip` so maybe we can improve that later.

hannyle:
> > is it possible to align vertically the question mark with the form field so that they look on the same line ? right now they look a bit out of place at least in my version of firefox
> 
> Easy one done in latest commit `replace help icon inline with textfield `. More difficult are the ones inside some box, did not found those nor the solution yet.

In `helpIconStyle > divBaseline`, I think we can leave out `flexWrap: wrap` so the ones inside the box can be on the same line.

lilachic:
Thanks, @hannyle , leaving out `flexWrap: wrap` made them inline inside box :) . Rest of styling can be fixed later when adding  CSC style, #363 .

Issue #351, opened 2021-06-22
OneOf field populates with previously filled options values


saulipurhonen:
#### Describe the bug

If oneOf option has saved data and user decides to change the option. Form tries to populate the new option with previous values. This happens only on #323 since form rendering doesn't work yet on other branches.

#### To Reproduce

Steps to reproduce the behaviour:

1. Go to Experiment form
2. Choose `Complex processing` from Expected Base Call Table > Processing
3. Fill Complex Processing fields
4. Save draft and open newly saved draft from draft object picker
5. Choose `Single processing` from Expected Base Call Table > Processing
6. Single Processing field has been tried to populate with object.

#### Expected behavior

Fields for new option should be clean.

PR #361, opened 2021-06-29
Fix for Processing OneOf fields in Experiment form

hannyle:
### Description

Fix for handling the `oneOf` field regarding to `Single processing` and `Complex processing` options in `Experiment` form.

### Related issues

Fix https://github.com/CSCfi/metadata-submitter-frontend/issues/351 and related issue https://github.com/CSCfi/metadata-submitter-frontend/pull/323#pullrequestreview-689512090

### Type of change

### Changes Made

- Prevent `FormTextField` to take input's `value` as an object
- `Unregister` input when selecting `Complex processing` so it won't parse the value of `Single processing` to its nested input fields

### Testing

- [x] Integration Tests

### DIFF
--- a/src/components/NewDraftWizard/WizardForms/WizardJSONSchemaParser.js
+++ b/src/components/NewDraftWizard/WizardForms/WizardJSONSchemaParser.js
@@ -347,13 +347,14 @@ const FormOneOfField = ({
         // Match key from currentObject to option property.
         // Field key can be deeply nested and therefore we need to have multiple cases for finding correct value.
         if (isNaN(parentPath[0])) {
-          fieldValue = (options.find(option => option.properties[parentPath])
-            ? // Eg. Sample > Sample Names > Sample Data Type
-              options.find(option => option.properties[parentPath])
-            : // Eg. Run > Run Type > Reference Alignment
-              options.find(
-                option => option.properties[Object.keys(flattenObject(itemValues))[0].split(".").slice(-1)[0]]
-              )
+          fieldValue = (
+            options.find(option => option.properties[parentPath])
+              ? // Eg. Sample > Sample Names > Sample Data Type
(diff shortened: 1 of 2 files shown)


blankdots (approved):
seems to fix the issue i had:https://github.com/CSCfi/metadata-submitter-frontend/pull/323#pullrequestreview-689512090 so approved

src/components/NewDraftWizard/WizardForms/WizardJSONSchemaParser.js
```diff
@@ -383,17 +384,14 @@ const FormOneOfField = ({
 
   if (itemValue) {
     switch (lastPathItem) {
-      case "prevStepIndex": {
-        fieldValue = "String value"
-      }
+      case "prevStepIndex":
```
lilachic, line 387: I wonder why there is switch instead of if or && when there is only one option for case?

lilachic (approved):
Looks good to me.

src/components/NewDraftWizard/WizardForms/WizardJSONSchemaParser.js
hannyle: This code is not part of this PR (actually from PR https://github.com/CSCfi/metadata-submitter-frontend/pull/323) but I can say that this is a temporary way of writing it in case we have some other fields that face the same error. If so, we can add more `case`. As explained in the comment also:
`// Special handling for Expected Base Call Table > Processing > Complex Processing > Pipeline > Pipe Section > Prev Step Index
// Can be extended to other fields if needed`

We can definitely shorten the code as the way you suggested if later there is no more `case` :)

lilachic: :) ok, thanks.

Issue #321, opened 2021-05-26
Draft object patch overwrite


saulipurhonen:
#### Describe the bug

Patching draft object doesn't overwrite values for fields that should have only one value. This can cause problems when populating form default values.

Overwrite works when submitting object so I think this might be related to backend.

#### To Reproduce

Steps to reproduce the behaviour:

1. Fill a `Sample` form with `Sample Data Type`
2. Save draft
3. Go to `Choose from drafts` and edit draft created in first step
4. Change `Sample Data type` from previous value and fill required field
5. Update draft
6. Go to `Choose from drafts` and edit draft created in first step
7. Inspect database for newly created draft object
8. See multiple key/value pairs in `sampleData`

#### Expected behavior

OneOf fields with only one possible value should be overwritten when draft object is patched

#### Screenshots

![image](https://user-images.githubusercontent.com/48789543/119641729-a026e080-be22-11eb-9588-14a0e872d649.png)

saulipurhonen:
This seems to work now.

saulipurhonen:
This problem happens on branch `bugfix/edit-does-not-work-for-all-fields`. It seems to be related to form populating and it keeps passes the old value on object patch actions. Also noticed that this does happen in both draft and submitted object scenarios.

Should this be handled in before patching or should backend catch this?

blankdots:
:thinking: will look into it and come back with some ideas, as not sure in fully understand if this is front or back issue

blankdots:
@saulipurhonen could I get an example of the `patch` requests made to the backend, will help me better debug this.

saulipurhonen:
@blankdots `PATCH` request payload for draft Sample:
`{"sampleName":{"taxonId":1234},"sampleData":{"gender":"male","dataDescription":"test description"},"status":"Draft"}`

saulipurhonen:
So there's both `gender` and `dataDescription` attributes. What I'm thinking is we have two options:

1) Keep only the latest one of and mutate the object in frontend before patching.
2) Backend compares payload to earlier data and makes mutation there.

blankdots:
the patch in backend does: https://docs.mongodb.com/manual/reference/operator/update/set/ ... might be that i need to pre-process the query to make in the format mongo wants

blankdots:
regarding this, I think it is combination of back-end and front-end issue, as front-end wants to do a PATCH but data is already there so it does the usual operation and tries to add stuff, which does not overwrite what is already there.
[...]

PR #417, opened 2021-08-09
 Fix for object before patching

hannyle:
### Description

`Sample Data Type` in **Sample** form is overwritten with only one value before the object is patched.
### Related issues

Fix https://github.com/CSCfi/metadata-submitter-frontend/issues/321

### Type of change

### Changes Made

- Remove the previous value (if exists) when selecting a new field of the same path in form 

### Testing

- [x] Integration Tests


blankdots (approved):
tested and data in db looks good

saulipurhonen (approved):
Looks good to me. Field is overwritten as supposed to.

PR #426, opened 2021-08-17
add retries & disable also video compression

blankdots:
### Description

Add cypress test retries and disable video compression

### Related issues

hopefully this makes cypress failures not so random

### Type of change

### Changes Made

1. add 3 retries for CI
2. remove videocompression to make CI run faster.

### Testing

- [x] Tests do not apply

### Mentions

retries **might** cause inconsistencies if data is not cleaned.

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -3,5 +3,7 @@
     "port": "3000",
     "mockAuthHost": "localhost",
     "mockAuthPort": "8000"
-  }
+  },
+  "retries": 2,
+  "videoCompression": false
 }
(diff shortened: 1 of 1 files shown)


saulipurhonen (approved):
LGTM

hannyle (approved):
I suppose setting `"videoCompression": false` also helps reduce the file size ?

blankdots:
> I suppose setting "videoCompression": false also helps reduce the file size ? 

actually does the opposite, there is no video compression at all

hannyle:
> > I suppose setting "videoCompression": false also helps reduce the file size ?
> 
> actually does the opposite, there is no video compression at all

Yep, I just checked from the document: `The compression step will be skipped completely, so the video will be large, but the processing should be faster.`

PR #453, opened 2021-09-15
Feature/json schema form tests

### LINKED ISSUE #54: Integration tests for JSON schema based forms
#### Proposed solution

Test that forms are rendered correctly.

#### DoD (Definition of Done)

Forms are tested in sensible way.

saulipurhonen:
### Description

Jest tests for form rendering from custom schema.

### Related issues

Closes #54 

### Type of change

### Changes Made

- Created custom schema with different field types
- Test form rendering

### Testing

- [x] Integration Tests


lilachic (approved):

hannyle (approved):
Looks nice to me! 

I'm not sure if we have covered all the cases in forms yet but at least I have seen most of them. 
I don't see the test case for `enum` in `jest` test for example, or is it already covered in cypress test (?)

And I think it's good also to have some comments (in `jest` test) about which case/form as an example to easier to catch up later.

saulipurhonen:
> I'm not sure if we have covered all the cases in forms yet but at least I have seen most of them.

There can be some underlying special cases but I think we can add those later if found.

> I don't see the test case for `enum` in `jest` test for example, or is it already covered in cypress test (?)

Good call, missed that one entirely. Introduced test for `enum` field just now.
 
> And I think it's good also to have some comments (in `jest` test) about which case/form as an example to be easier to catch up later.

Very good point. Added some comments to latest commit.

hannyle (approved):
I think we have covered most of the cases for now. 
If later any specific cases still need to be added, we could have them in another PR.

PR #474, opened 2021-10-04
Test form validation

### LINKED ISSUE #147: Test JSON schema based form validation
#### Proposed solution

Form validations should be tested. Testing should be done in Cypress.

#### DoD (Definition of Done)

Validations are properly tested.

saulipurhonen:
### Description

Test that form validates and doesn't submit with invalid form.

### Related issues

Closes #147 

### Type of change

### Changes Made

- Test that typed field validates and renders hint message
- Test that invalid form doesn't submit
- Added longer timeout for autocomplete field test. Default timeout caused issues occasionally

### Testing

- [x] Integration Tests
- [x] E2E

### DIFF
--- a/cypress/integration/app.spec.js
+++ b/cypress/integration/app.spec.js
@@ -21,6 +21,12 @@ describe("Basic e2e", function () {
     cy.get("div[role=button]").contains("Fill Form").type("{enter}")
     cy.get("div[role=button]").contains("Skip to form")
 
+    // Try to send invalid form
+    cy.get("button[type=submit]").contains("Submit").click()
+    cy.get("input[name='descriptor.studyTitle']").then($input => {
+      expect($input[0].validationMessage).to.eq("Please fill out this field.")
+    })
+
     // Fill a Study form and submit object
     cy.get("input[name='descriptor.studyTitle']").type("Test title")
     cy.get("button[type=button]").contains("Clear form").click()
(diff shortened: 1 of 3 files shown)


blankdots (approved):

hannyle:
The integration test looks good to me, but I think we might need to test the validation on `submit` also ? Because some forms does not validate on `blur`. 
For example, in DAC form, `Contacts` is a required field and it only shows error on `submit`.

saulipurhonen:
> 
> 
> The integration test looks good to me, but I think we might need to test the validation on `submit` also ? Because some forms does not validate on `blur`. For example, in DAC form, `Contacts` is a required field and it only shows error on `submit`.

Trying to submit a DAC form without adding contact causes error and can't be tested right now.
I opened a bug issue for this: #477  

`Main Contact` field validation is browser based required field behavior and is tested in E2E test in `app.spec.js`.

hannyle:
> > The integration test looks good to me, but I think we might need to test the validation on `submit` also ? Because some forms does not validate on `blur`. For example, in DAC form, `Contacts` is a required field and it only shows error on `submit`.
> 
> Trying to submit a DAC form without adding contact causes error and can't be tested right now. I opened a bug issue for this: #477
> 
> `Main Contact` field validation is browser based required field behavior and is tested in E2E test in `app.spec.js`.

okay, I might be confused with this form validation test purpose. If I understood correctly, we are testing the field validation in form only.

saulipurhonen:
> okay, I might be confused with this form validation test purpose. If I understood correctly, we are testing the field validation in form only.

PR title could be more accurate since the original issue is titled as `Test JSON schema based form validation`. So in this case, form only.

Issue #363, opened 2021-06-29
Switch to CSC colour scheme and styling


lilachic:
#### Description

CSC UX group provides CSC design system, we would like to to switch our UX according to that.

CSC styling system should bring along CSC colours, SCS styling  and accessibility.

for now we are not going to use it due to missing components:
> CSC design system:
> https://cscui.a3s.fi/index.html 

Tips about accessibility:
https://webaim.org/articles/contrast/
https://wave.webaim.org/extension/

Relates to #170 and #105 as other design issues.

#### DoD (Definition of Done)
When the UX colours and styling mainly conforms CSC design  system, accessibility checked.

#### Testing
Peer review.

lilachic:
Help icons, `helpIconStyle` , question marks at forms need to be improved in terms of accessibility, contrast.

blankdots:
the lack of UI components and accessibility and we cannot ahead and use the CSC design system.

hannyle:
Here I got the latest color guidelines for the UI.
- Background color: Darkgreen 10%

[CSC Design system + instructions.pdf](https://github.com/CSCfi/metadata-submitter-frontend/files/7525385/CSC.Design.system.%2B.instructions.pdf)

PR #475, opened 2021-10-05
Feature/using draft templates

hannyle:
### Type of change

- [ ] New feature (non-breaking change which adds functionality)

### Testing

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Tests do not apply


blankdots (approved):
Tested with using one and multiple templates in a submission, and advertised behaviour is there: A user can use templates in a submission.

Not sure if this is a problem, but if a user selects a template for a submission and does not fill that draft, it might end up saving it as template, thus resulting in multiple duplicate templates - This might be a non-issue, so let us see how it will turn out in practice.

hannyle:
> Tested with using one and multiple templates in a submission, and advertised behaviour is there: A user can use templates in a submission.
> 
> Not sure if this is a problem, but if a user selects a template for a submission and does not fill that draft, it might end up saving it as template, thus resulting in multiple duplicate templates - This might be a non-issue, so let us see how it will turn out in practice.

I understand. This relates to user experience so yes, we'll see.

saulipurhonen:
Tested and works as supposed.

What could be improved (on other PR) is to separate templates from drafts. This could fix the UX issue @blankdots mentioned.
In this scenario user checks templates the user wants to use. Templates would appear as separate list at the sidebar with submitted / draft objects.

blankdots:
> Tested and works as supposed.
> 
> What could be improved (on other PR) is to separate templates from drafts. This could fix the UX issue @blankdots mentioned.
> In this scenario user checks templates the user wants to use. Templates would appear as separate list at the sidebar with submitted / draft objects.

indeed a good idea, however i suggest to wait on this change/issue till #363 is started to be sure we don't do the work twice.

hannyle:
> > Tested and works as supposed.
> > What could be improved (on other PR) is to separate templates from drafts. This could fix the UX issue @blankdots mentioned.
> > In this scenario user checks templates the user wants to use. Templates would appear as separate list at the sidebar with submitted / draft objects.
> 
> indeed a good idea, however i suggest to wait on this change/issue till #363 is started to be sure we don't do the work twice.

I agree. In fact there are some proposed solutions for highlighting templates from drafts in the new UI (WIP) as well. We shall see when we start to do the UI improvement.

PR #482, opened 2021-10-07
i18n support

### LINKED ISSUE #21: Add i18n internationalization support
#### Description

Add internationalization support using https://react.i18next.com/ (or similar) for Finnish and English.

#### DoD (Definition of Done)

UI is able to switch between language seamlessly.
There is a dropdown that allows language switching in the menu bar 

#### Testing

Peer review (for now)

saulipurhonen:
### Description

Introduced i18n support with [react-i18next library](https://react.i18next.com/).
Selection is stored in local storage. This helps to render correct translation when user chooses locale on login page.

### Related issues

Closes #21 

### Type of change

### Changes Made

- Added i18n configuration file to `src` path
- Language selector on top navigation with MUI Menu component
- Store locale selection to browsers local storage and redux state
- Utility function for navigating with selected locale
- Changed navigation items to navigate with locale
- Added translation files to `src/translations`
- Translated "Logged in as" string on `/home` route. 
- Added fallback for user templates in `UserDraftTemplates`. This caused some issues if templates resulted as undefined

### Testing

- [x] E2E

### DIFF
--- a/src/__tests__/WizardObjectDetails.test.js
+++ b/src/__tests__/WizardObjectDetails.test.js
@@ -2,8 +2,7 @@ import React from "react"
 
 import "@testing-library/jest-dom/extend-expect"
 import { ThemeProvider } from "@material-ui/core/styles"
-import { render, screen, waitForElementToBeRemoved } from "@testing-library/react"
-import userEvent from "@testing-library/user-event"
+import { render, screen } from "@testing-library/react"
 import { rest } from "msw"
 import { setupServer } from "msw/node"
 import { Provider } from "react-redux"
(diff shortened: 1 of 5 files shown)


saulipurhonen:
Issue #21 includes feature requests for both JSON schema based forms and AJV validation. 
To reduce weight on this PR I think we should do these two on separate PR's.

saulipurhonen:
> 
> 
> For me the components inside `<NewDraftWizard/>` seem not completely worked with translation. For example, `WizardCreateFolderStep` is missing the query params when changing language.

Query parameters are now included in language change method.

Issue #460, opened 2021-09-28
Replacing XML file with the same name gives error


hannyle:
#### Describe the bug

Replacing XML file with the same name gives error

#### To Reproduce

Steps to reproduce the behaviour:

1. Go to 'Study' 
2. Select 'Upload XML File' and click `Submit`
3. Replace the submitted file with the same file name

#### Expected behavior

Submitting the same file name should be possible.

#### Screenshots

![image](https://user-images.githubusercontent.com/23532204/135080142-a793e047-1374-4574-86be-0f8cfb4eb6a2.png)

blankdots:
> However, the same XML file name issue is not resolved yet. Could this be fixed in backend or do you have any suggestions ?

i think it is fine how it is solved in #483, apparently the issue was replacing forbidden fields and we now show an error for it.
So the issues is solved with that PR.

PR #483, opened 2021-10-08
Catch and show error message when replacing the same XML file

hannyle:
### Description

Error when replacing XML file with the same name will not be redirected to page error400 but shown as an error message.

### Related issues

Solves : https://github.com/CSCfi/metadata-submitter-frontend/issues/460

### Type of change

### Changes Made

- Add exceptional case in `errorMonitor.js` for handling error when replacing XML file

### Testing

- [x] Tests do not apply

### DIFF
--- a/cypress/integration/app.spec.js
+++ b/cypress/integration/app.spec.js
@@ -52,7 +52,23 @@ describe("Basic e2e", function () {
     // Saved objects list should have newly added item from Study object
     cy.get(".MuiListItem-container", { timeout: 10000 }).should("have.length", 2)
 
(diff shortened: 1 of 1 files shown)


blankdots:
> Looks good to me! One future enhancement could be to improve error messages. In this particular case user might want to know what fields of the file are causing the problem.

yes. we can create an issue about this. In backend we mostly follow https://datatracker.ietf.org/doc/html/rfc7807#section-3.1 to expose errors. Most of JSON/XML validation errors are 400 and login errors 401/403.

PR #522, opened 2021-11-04
subject scheme should be fixed

blankdots:
### Description

forgot `subjectscheme` should be fixed. Will also remove it in backend

### Type of change

### Changes Made

1. auto add subject scheme to all subjects

### Testing

- [x] Tests do not apply


hannyle (approved):
Looks good to me!

PR #528, opened 2021-11-05
Feature/enable draft update

### LINKED ISSUE #488: Users can Update templates
#### Description

With #247 we introduce option to add templates, in order to keep the list of templates updated, provide a mechanism for a user to update the template. `PATCH` is available in backend.

#### DoD (Definition of Done)

Users can update data in a template

#### Testing

Unit and Integration testing.

saulipurhonen:
### Description

Added option for user to update saved templates.

### Related issues

Closes #488

### Type of change

### Changes Made

- Template menu action for edit functionality
- Simplified templates list rendering (previous version caused nested <form> tags with form dialog)
- Option for rendering `WizardFillObjectDetailsForm` component in a dialog
- Form actions for template -typed object
- Reset draftStatus and currentObject redux state objects on form component unmount
- Handle template replace both into backend and redux state
- Prevent re-render of app by moving status message handling rendering logic to message handler from app root
- Serialize status message details
- Updated status message handler tests
- Added test for template edit

### Testing

- [x] Integration Tests

### DIFF
--- a/cypress/integration/draftTemplates.spec.js
+++ b/cypress/integration/draftTemplates.spec.js
@@ -97,7 +97,7 @@ describe("draft selections and templates", function () {
       cy.get("h1", { timeout: 10000 }).contains("Sample").should("be.visible")
       cy.get("[data-testid='title']").should("have.value", "Sample draft title")
     }),
-    it("should be able to select and delete draft templates and reuse them when creating a new folder", () => {
+    it("should be able to select, edit and delete draft templates and reuse them when creating a new folder", () => {
       // Select drafts inside the dialog
       cy.get("form").within(() => {
         cy.get("input[type='checkbox']").first().check()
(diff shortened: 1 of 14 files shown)


saulipurhonen:
Changes to status message handling had to be made since re-rendering whole app when status message state changes would unmount certain elements, eg. dialogs.

hannyle (approved):
Looks good to me!

hannyle:
I've actually encountered this error from `Console` when updating a template

Screenshot 2021-11-09 at 11 03 13

saulipurhonen:
> I've actually encountered this error from `Console` when updating a template

Good call. This error was caused because I typed response as optional property for `StatusDetails` type. Actions need to be serialized and this caused problems when response has nested unserialized data.

Fixed by stringifying response.

blankdots (approved):
image

Looks good, just a few notes:
- could be a tad wider 
- isn't clear form + update the same as delete. my question would be what would be the use case for `clear form` ? - we could also omit it

saulipurhonen:
> * could be a tad wider
> * isn't clear form + update the same as delete. my question would be what would be the use case for `clear form` ? - we could also omit it

Good points. I think we could change the width once we're doing whole design overhaul.

Not sure if "Clear form" button is needed. Could there be some use cases where user would want to use the template id but re-fill the form?

May be we could prevent updating empty form?

blankdots:
> Good points. I think we could change the width once we're doing whole design overhaul.

ok

> Not sure if "Clear form" button is needed. Could there be some use cases where user would want to use the template id but re-fill the form?
> 
> May be we could prevent updating empty form?

let us remove `clear form`, I don't see a use case for it, and if we find one we will add it back
And yes we should prevent empty form, it there is the case, but do you mean prevent in front-end or back-end ?

saulipurhonen:
> let us remove `clear form`, I don't see a use case for it, and if we find one we will add it back And yes we should prevent empty form, it there is the case, but do you mean prevent in front-end or back-end ?

I mean preventing empty form in front-end which is in the matter of fact implemented already.

Issue #542, opened 2021-11-22
Migrating Material UI from v.4 to v.5


hannyle:
#### Proposed solution

Migrating Material UI from v.4 to v.5

#### DoD (Definition of Done)

Material UI is upgraded to v.5 and the application should run without issues.

#### Testing

[x] Unit tests and Integration tests

PR #543, opened 2021-11-25
Feature/upgrade to Material-UI version 5

hannyle:
### Description

Update Material-ui from v.4 to v.5

### Related issues

Closes https://github.com/CSCfi/metadata-submitter-frontend/issues/542

### Type of change

### Changes Made

- The changes follow these steps: https://mui.com/guides/migration-v4/#update-mui-version
- Main packages are renamed from:

> @material-ui/core to @mui/material
> @material-ui/lab to @mui/lab
> @material-ui/icons to @mui/icons-material

- Related codes in all files are updated by using `codemod`: `npx @mui/codemod v5.0.0/preset-safe <path | folder>`
- *Important*: 
   - MUI v.5 uses [emotion](https://github.com/emotion-js/emotion) as the style library and the support for JSS utility like `makeStyles` and `withStyles` will be deprecated from the next version (still in discussion). MUI v.5 recommends other ways to style the components which are `styled` or `sx`. More info here: https://mui.com/customization/how-to-customize/
   - Since we are in the slow progress of switching to the new UI, I would suggest not to apply the new ways of styling (using `sx` or `styled`) for all the files now, instead, making changes when we implement each new UI's component/feature. This way we could avoid time consuming and the risk of bugs.
   - Document for new MUI will need to be updated afterwards.

### Testing

- [x] Unit Tests
- [x] Integration Tests


lilachic:
Regarding the header "Changes Made" Is @mui/lab in use at the moment? At least Alert is at "@mui/material/Alert" older version got path "@material-ui/lab/Alert" and  with "@mui/material/Autocomplete" older was "@material-ui/lab/Autocomplete".

src/__tests__/WizardStepper.test.js
```diff
@@ -21,7 +23,16 @@ describe("WizardStepper", () => {
       <MemoryRouter initialEntries={[{ pathname: "/newdraft", search: "?step=0" }]}>
         <Provider store={store}>
           <Routes>
-            <Route path="/newdraft" element={<WizardStepper />} />
+            <Route
+              path="/newdraft"
+              element={
+                <StyledEngineProvider injectFirst>
+                  <ThemeProvider theme={CSCtheme}>
+                    <WizardStepper />{" "}
```
saulipurhonen, line 31:
```suggestion
                    <WizardStepper />
```

src/components/Nav.js
```diff
@@ -49,9 +49,6 @@ const useStyles = makeStyles(theme => ({
   linkButton: {
     margin: theme.spacing(1, 1.5),
   },
-  toolbar: {
-    flexWrap: "wrap",
```
saulipurhonen, line 53: Flex wrap seems to work better on current layout in mobile screen widths. However changes to navigation and collapsible navigation will need rework so I think it's ok no remove wrapping for now.

saulipurhonen:
Old stepper:
image

MUI 5 stepper:
image

I'm not sure if it makes any sense to try to fix these to match original styles at this point as we're going to do whole design overhaul?

lilachic:
In uploading file the text is white and the "Submit" button is white.

saulipurhonen (approved):
LGTM!  

I think it might be unnecessary to fix design related issues (if issue doesn't prevent developing new features) before doing the new layout.

Also to mention, ForwardRef implementation on statusMessageHandler is a nice solution.

lilachic (approved):
LGTM

rquazi (approved):
Though I can't say I understood everything, but looks fine to me also .

src/components/Nav.js
hannyle: I removed this because for some reasons with the new MUI, flexWrap is kind of covering other elements so some cypress tests have failed because of that.

blankdots:
#544 is  needed for tests to pass

hannyle:
> Regarding the header "Changes Made" Is @mui/lab in use at the moment? At least Alert is at "@mui/material/Alert" older version got path "@material-ui/lab/Alert" and with "@mui/material/Autocomplete" older was "@material-ui/lab/Autocomplete".

Good point! With the new MUI, `Alert` and `Autocomplete` have been moved from "@mui/material", but we still need the `@mui/lab`. One of my developing branch is using it.

hannyle:
> Old stepper: image
> 
> MUI 5 stepper: image
> 
> I'm not sure if it makes any sense to try to fix these to match original styles at this point as we're going to do whole design overhaul?




> I think it might be unnecessary to fix design related issues (if issue doesn't prevent developing new features) before doing the new layout.
>

--> Yes, I think it would be better to fix the design related issues while doing the new UI as it would probably change totally.

PR #560, opened 2021-12-03
Flow to TypeScript conversion

### LINKED ISSUE #448: Flow to TypeScript conversion
#### Proposed solution

TypeScript seems to be superior to Flow. Facebook has announced that they're focusing Flow more heavily into their own products and this might cause problems in the future if we'd stick with Flow.

There are some Flow to TypeScript converters which we can try out.

Two most popular repos are:
- [Khan Flow-to-ts](https://github.com/Khan/flow-to-ts)
- [Flow-to-TypeScript](https://github.com/bcherny/flow-to-typescript)

#### DoD (Definition of Done)

- Typings are converted to TS syntax
- Husky should lint files with TypeScript

#### Testing

Linting with TS works

saulipurhonen:
### Description

Project wide conversion from Flow to TypeScript.

### Related issues

Closes #448 

### Type of change

### Changes Made

- TS configurations (in eslintrc.json and tsconfig.json)
- ESLint configurations with TS
- Converted files to .ts and .tsx extensions
- Installed necessary types
- Use redux useSelector and useDispatch as custom typed methods (useAppSelector, useAppDispatch)
- Module augmentations for MUI Theme and Palette to enable custom properties
- Improved types throughout project tree
- Removed Flow related configurations, including workflow for static types check. This is handled via eslint
- Cypress TS configuration
- Introduced types for custom Cypress commands
- Updated README.md
- Introduced TSC check in Husky pre-commit hook and GitHub Actions workflow

### Testing

- [x] Tests do not apply


blankdots (commented):
a comment while i review this, is it possible or useful to use the tsc (https://www.typescriptlang.org/docs/handbook/compiler-options.html) to do typechecks e.g. https://github.com/typescript-eslint/typescript-eslint/blob/main/package.json#L48 so that we can implement that as part of a github action ?

saulipurhonen:
> a comment while i review this, is it possible or useful to use the tsc (https://www.typescriptlang.org/docs/handbook/compiler-options.html) to do typechecks e.g. https://github.com/typescript-eslint/typescript-eslint/blob/main/package.json#L48 so that we can implement that as part of a github action ?

I think this a great idea and in addition should be added to Husky pre-commit hooks. I'll try to suggest both workflow and GH action workflow into this PR.

saulipurhonen:
> > a comment while i review this, is it possible or useful to use the tsc (https://www.typescriptlang.org/docs/handbook/compiler-options.html) to do typechecks e.g. https://github.com/typescript-eslint/typescript-eslint/blob/main/package.json#L48 so that we can implement that as part of a github action ?
> 
> I think this a great idea and in addition should be added to Husky pre-commit hooks. I'll try to suggest both ~workflow~ pre-commit hook and GH action workflow into this PR.

Husky pre-commit hook and GH action workflow now introduced.

hannyle (approved):
Looks very good to me!

I still have a question: Why do we need to put `dispatch` as `[dispatch]` in `useEffect`, for e.g. in `App.tsx`, instead of `[ ]` as usual ? Any documents regarding to this ?

saulipurhonen:
> I still have a question: Why do we need to put `dispatch` as `[dispatch]` in `useEffect`, for e.g. in `App.tsx`, instead of `[ ]` as usual ? Any documents regarding to this ?

Good call.

Empty second arguments on useEffect Hook is against `react-hooks/exhaustive-deps` eslint rule. In recommended eslint configuration we'd have this rule to warn on improperly used hooks.

I had this rule configured at the beginning of conversion but I decided to leave it off at the moment to help finish the conversion faster.

.github/workflows/typecheck.yml
```diff
@@ -8,16 +8,16 @@ jobs:
     strategy:
       matrix:
         node: ["14"]
-    name: Flow check
+    name: TypeScript check
     steps:
       - uses: actions/checkout@v2
       - name: Setup node ${{ matrix.node }}
-        uses: actions/setup-node@v2.5.0
+        uses: actions/setup-node@v2.4.0
```
blankdots, line 15: why did we downgrade the version here?

README.md
```diff
@@ -2,7 +2,6 @@
 
 ![Unit tests and coverage](https://github.com/CSCfi/metadata-submitter-frontend/workflows/Unit%20tests%20and%20coverage/badge.svg)
 ![Code style check](https://github.com/CSCfi/metadata-submitter-frontend/workflows/Code%20style%20check/badge.svg)
-![Static type check](https://github.com/CSCfi/metadata-submitter-frontend/workflows/Static%20type%20check/badge.svg)
```
blankdots, line 5: maybe we can reintroduce the badge, no need to delete it now

blankdots (approved):
there are a lot of types as `any`, but I assume we can revisit that.

saulipurhonen:
> there are a lot of types as `any`, but I assume we can revisit that.

This was on purpose to save time. I think we could introduce eslint rule `no-explicit-any` and type all instances of `any`. This however can be time consuming and I suggest this to be done on different PR.

README.md
saulipurhonen: Sounds good to me.

.github/workflows/typecheck.yml
saulipurhonen: This seems to be a mistake and should be v2.5.0.

blankdots:
> This was on purpose to save time. I think we could introduce eslint rule `no-explicit-any` and type all instances of `any`. This however can be time consuming and I suggest this to be done on different PR.

that what is what i implied as well.

also sry did not realised me merging the dependencies cause package lock to get conflicts

saulipurhonen:
> also sry did not realised me merging the dependencies cause package lock to get conflicts

No prob.

blankdots:
I continued on this PR as established wiht @saulipurhonen in RC

blankdots:
seems pressing the `Submit Study` button does not work for some reason :man_shrugging:

src/components/Nav.tsx
```diff
@@ -10,7 +10,7 @@ import MenuItem from "@mui/material/MenuItem"
 import Toolbar from "@mui/material/Toolbar"
 import Typography from "@mui/material/Typography"
 import { makeStyles } from "@mui/styles"
-import i18n from "i18next"
+import { default as i18n } from 'i18next'
```
blankdots, line 13: i don't know why this work.
I followed this guide: https://github.com/i18next/i18next/issues/1692

hannyle:
> ~seems pressing the `Submit Study` button does not work for some reason 🤷‍♂️~
> 
> `FormSelectField` seem to be broken

--> Fixed for `FormSelectField`, the function responsible for changing fieldValue was broken in this branch

The PR is in quite a good shape now, there are still some issues left as mentioned: types as `any`, and enable eslint rule `react-hooks/exhaustive-deps` again, but I think this could be done separately in another PR.

src/components/Nav.tsx
hannyle: Eslint was so strict with this rule: [no-named-as-default-member](https://github.com/import-js/eslint-plugin-import/blob/v2.25.3/docs/rules/no-named-as-default-member.md) 

The module has `i18n` as default export and other export properties like `use`, `changeLanguage`. According to eslint, we need separate them as <`import i18n, {use, changeLanguage} from "i18next"`> or in this case, I just import everything from the module: <`import * as i18n from "i18next"`>

blankdots:
then I will click merge and we can see what happens.

Issue #531, opened 2021-11-08
New login page


blankdots:
#### Description

as part of move to new design styles #363 we need a new UI page
https://www.figma.com/file/NJiSfMKCPMJZIduAYVq0he/SD-vs-2?node-id=0:1

@hannyle to provide image

example page also from https://sd-desktop.csc.fi

#### DoD (Definition of Done)

New login page fits the design

#### Testing

Peer review

hannyle:
![csc_banner_login](https://user-images.githubusercontent.com/23532204/140871004-63bd7923-ba1a-4c7d-b207-43b1e30a5e9a.png)

PR #573, opened 2021-12-13
Feature/new login page

hannyle:
### Description

- Introducing new Login page based on the design and switching to CSC Color scheme and font styles.
- Updating new components with MUI v.5 and Typescript

### Related issues

Closes https://github.com/CSCfi/metadata-submitter-frontend/issues/363 and https://github.com/CSCfi/metadata-submitter-frontend/issues/531

### Type of change

### Changes Made

- Switched our current color theme to CSC color system and update its `typescript` file `theme.d.ts`
- Imported the new font `Museo Sans` to match with CSC Design system and declare in `typescript` file `font.d.ts`
- Rewrote the whole Login page and some parts of Navigation bar (for Login page) according to the new UI design, with new version of MUI 5 and `typescript`
- Updated cypress tests

### Testing

- [x] Integration Tests


blankdots (approved):

lilachic (approved):
LGTM. Noticed that with narrow window the banner image is cut from right side thus it does not fill the container all the way to the top at at right side
Screenshot 2021-12-28 at 10 21 35
.

rquazi (approved):
looks good to me :)

hannyle:
> LGTM. Noticed that with narrow window the banner image is cut from right side thus it does not fill the container all the way to the top at at right side Screenshot 2021-12-28 at 10 21 35 .

This is actually more about how we decide the UI would look like on different screens. To have the exact look the best way would be having different banner images for different screen sizes, but I have made a temporary fix for now, we may come up with a better solution later.

PR #621, opened 2022-01-03
Use React Hooks by ESLint hooks rule

### LINKED ISSUE #561: Incorrect use of hooks in callback functions
#### Describe the bug

It seems that we're using React Hooks in incorrect way. We have multiple cases where Hooks are used inside of callback functions.

#### To Reproduce

1. Add `"react-hooks/rules-of-hooks": "error"` into eslint -config file - rules section
2. See eg. `WizardJSONSchemaParser` file for errors

#### Expected behavior

React Hooks must be called in a React function component or a custom React Hook function.

saulipurhonen:
### Description

Introduced ESLint rule for use of React Hooks. This rule ensures hooks are always called in same exact order and should prevent possible memory leaks.

### Related issues

Closes #561 

### Type of change

### Changes Made

- Added rules-of-hooks rule to ESLint configuration
- Added static type for templates
- Transformed callback and conditional Hook calls to fit Hooks guidelines (see [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html))

### Testing

- [x] Tests do not apply

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -54,6 +54,7 @@
       }
     ],
     "no-unused-vars": "off",
+    "react-hooks/rules-of-hooks": "error",
     "@typescript-eslint/no-unused-vars": "error",
     "@typescript-eslint/no-explicit-any": ["off"],
     "@typescript-eslint/explicit-module-boundary-types": "off"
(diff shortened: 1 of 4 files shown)


hannyle (approved):
LGTM! It's good we could enable this rule now.

PR #624, opened 2022-01-14
Feature/disallow any type

### LINKED ISSUE #619: Disallow usage of the any type
#### Proposed solution

Disallow any type to keep typing more efficient.

#### DoD (Definition of Done)

Linting passes when any type is disallowed in eslint rules.

#### Testing

 `npm run lint` lints without errors.

saulipurhonen:
### Description

Converted "any" typed properties and attributes to corresponding types.
Prevent use of "any" type with ESLint & TSConfig rules.

### Related issues

Closes #619 

### Type of change

### Changes Made

- Added ESLint rule for disallowing "any" type
- Configured TSConfig to disallow "any" type
- Increased default Cypress timeout value to 10000. Fixes typing problems on commands that are typed via jQuery types.
- Adjusted schema fetching logic (eg. in `src/components/NewDraftWizard/WizardForms/WizardDOIForm.tsx` to work with stricter typing
- Introduced new static types, eg: ConnectFormMethods, FormObject...
- Introduced CHANGELOG.md

### Testing

- [x] Tests do not apply


saulipurhonen:
There are still some occasions of "any" type that are handled by bypassing rules.

1. `src/rootReducer.ts`:
   - Root reducer has "any" as return type. Removing this type reveals possible typing problems at least with `currentObject` state object. This problem can be seen in `WizardForms/WizardJSONSchemaParser.tsx` if any type is removed from root reducer.

2. `WizardForms/WizardJSONSchemaParser.tsx`:
   - Argument for `traverseFormValuesForCleanUp` has "any" type.

3. `cypress/integration/linkingAccessionIds.spec.ts`:
   - Whole Cypress test file is allowed for "any" type. I think usage of variables need to be written differently. See [sharing context](https://docs.cypress.io/guides/core-concepts/variables-and-aliases#Sharing-Context).

blankdots (approved):
looks way better now in terms of types.

saulipurhonen:
> There are still some occasions of "any" type that are handled by bypassing rules.
> 
> 1. `src/rootReducer.ts`:
>    
>    * Root reducer has "any" as return type. Removing this type reveals possible typing problems at least with `currentObject` state object. This problem can be seen in `WizardForms/WizardJSONSchemaParser.tsx` if any type is removed from root reducer.
> 2. `WizardForms/WizardJSONSchemaParser.tsx`:
>    
>    * Argument for `traverseFormValuesForCleanUp` has "any" type.
> 3. `cypress/integration/linkingAccessionIds.spec.ts`:
>    
>    * Whole Cypress test file is allowed for "any" type. I think usage of variables need to be written differently. See [sharing context](https://docs.cypress.io/guides/core-concepts/variables-and-aliases#Sharing-Context).

These are fixed.

saulipurhonen:
> Also it seems like TypeScript is trying to check libraries in node modules folder while running E2E tests in GitHub Actions.
> 
> It gives following error:
> 
> ```
> TypeScript error in /home/runner/work/metadata-submitter-frontend/metadata-submitter-frontend/src/index.tsx(4,22):
> Could not find a declaration file for module 'react-dom'. '/home/runner/work/metadata-submitter-frontend/metadata-submitter-frontend/node_modules/react-dom/index.js' implicitly has an 'any' type.
>   Try `npm i --save-dev @types/react-dom` if it exists or add a new declaration (.d.ts) file containing `declare module 'react-dom';`  TS7016
> ```

This is also fixed. `"noImplicitAny": false,` in tsconfig doesn't validate node modules.

lilachic (approved):
LGTM.

hannyle:
The types are more elaborately declared and look really good to me.

cypress/integration/app.spec.ts
```diff
@@ -25,7 +27,7 @@ describe("Basic e2e", function () {
     // Try to send invalid form
     cy.formActions("Submit")
 
-    cy.get("[data-testid='descriptor.studyTitle']").then(($input: any) => {
+    cy.get<FormInput[]>("input[name='descriptor.studyTitle']").then($input => {
```
hannyle, line 30:
```suggestion
cy.get<FormInput[]>("input[data-testid='descriptor.studyTitle']").then($input => {
```

src/components/Home/SubmissionFolderList.tsx
```diff
@@ -144,16 +144,18 @@ const SubmissionFolderList: React.FC<any> = () => {
...
-        folderType={location === "drafts" ? FolderSubmissionStatus.unpublished : FolderSubmissionStatus.published}
-        folders={location === "drafts" ? unpublishedFolders : publishedFolders}
-        location={location}
-        page={page}
-        itemsPerPage={itemsPerPage}
-        totalItems={location === "drafts" ? totalFolders.totalUnpublishedFolders : totalFolders.totalPublishedFolders}
-        fetchItemsPerPage={handleFetchItemsPerPage}
-        fetchPageOnChange={handleFetchPageOnChange}
-      />
+      {
```
hannyle, line 147: Is there an extra `{`  ?

src/components/NewDraftWizard/WizardForms/WizardDOIForm.tsx
```diff
@@ -15,38 +15,45 @@ import { resetCurrentObject } from "features/wizardCurrentObjectSlice"
...
     const getDataciteSchema = async () => {
       let dataciteSchema = sessionStorage.getItem(`cached_datacite_schema`)
+      let parsedDataciteSchema: FormObject | undefined = undefined
+
       if (!dataciteSchema || !new Ajv().validateSchema(JSON.parse(dataciteSchema))) {
         try {
           const response = await schemaAPIService.getSchemaByObjectType("datacite")
-          dataciteSchema = response.data
+          dataciteSchema = JSON.stringify(response.data)
           sessionStorage.setItem(`cached_datacite_schema`, JSON.stringify(dataciteSchema))
```
hannyle, line 39: We have `JSON.stringify` twice

src/components/NewDraftWizard/WizardHooks/WizardTransformTemplatesToDrafts.tsx
```diff
@@ -8,7 +8,7 @@ import { getObjectDisplayTitle, getOrigObjectType } from "utils"
 const transformTemplatesToDrafts = async (
   templateAccessionIds: Array<string>,
   templates: Array<ObjectInsideFolder>,
-  dispatch: (dispatch: (reducer: any) => void) => void
+  dispatch: (dispatch: (reducer: never) => void) => void
```
hannyle, line 11: Having type `never` for `reducer` is a bit unclear to me. Could `type { AppDispatch }` be possible in this case ?

src/components/Home/SubmissionFolderList.tsx
saulipurhonen: Good catch, this seems to be unnecessary.

saulipurhonen: Change is pushed to the branch.

src/components/NewDraftWizard/WizardForms/WizardDOIForm.tsx
saulipurhonen: Latest changes has only one on variable set.

src/components/NewDraftWizard/WizardHooks/WizardTransformTemplatesToDrafts.tsx
saulipurhonen: Good catch. `AppDispatch` type suits it a lot better.

hannyle:
I could see that we have both `index signatures` and `Record` types. They share some similarities and differences as far as I know. In this example,
`WizardShowSummaryStep.tsx` 
```
{groupedObjects.map((group: Record<string, ObjectInsideFolderWithTags[]>) => {
```
vs `WizardDraftSelections.tsx`

```
{draftObjects.map((draft: { [x: string]: ObjectInsideFolderWithTags[] }) => {
```

In this case`groupedObjects` and `draftObjects` seem to have similar structure to me but have different types. Do you have any preferences when to use what?

saulipurhonen:
> I could see that we have both `index signatures` and `Record` types. They share some similarities and differences as far as I know. In this example, `WizardShowSummaryStep.tsx`
> 
> ```
> {groupedObjects.map((group: Record<string, ObjectInsideFolderWithTags[]>) => {
> ```
> 
> vs `WizardDraftSelections.tsx`
> 
> ```
> {draftObjects.map((draft: { [x: string]: ObjectInsideFolderWithTags[] }) => {
> ```
> 
> In this case`groupedObjects` and `draftObjects` seem to have similar structure to me but have different types. Do you have any preferences when to use what?

Good catch, what I've read is that both of these types are basically equivalent ([some corner cases exist](https://stackoverflow.com/a/54100359/13164589)). In some cases typing with index signatures might be a bit more informative for developer.

I don't personally have preference. However in the example file I'd like to have a coherent and shared type for mapping of both `groupedObjects` and `draftObjects`.

@hannyle do you think if we should decide to use either in the whole project or would it suffice that we mix these two different typing methods depending on the use case?

saulipurhonen:
`WizardShowSummaryStep.tsx`

> {groupedObjects.map((group: Record<string, ObjectInsideFolderWithTags[]>) => {

 `WizardDraftSelections.tsx`

> {draftObjects.map((draft: { [x: string]: ObjectInsideFolderWithTags[] }) => {

I added a new type for lists of objects by schema. `Record<string, ObjectInsideFolderWithTags[]>`

These both share this type now.

hannyle:
> Good catch, what I've read is that both of these types are basically equivalent ([some corner cases exist](https://stackoverflow.com/a/54100359/13164589)). In some cases typing with index signatures might be a bit more informative for developer.

> I don't personally have preference. However in the example file I'd like to have a coherent and shared type for mapping of both groupedObjects and draftObjects.
> @hannyle do you think if we should decide to use either in the whole project or would it suffice that we mix these two different typing methods depending on the use case?

I think there are still debations on which one is better or using both, but  in our case, I think it's fine to leave them mixed as they are now. 
As you also mentioned, `index signature` probably a better one to understand, and we may use `Record` in more specific cases if we know the `key` in advance [(well explained here)](https://dmitripavlutin.com/typescript-index-signatures/). We may take them into consideration for future cases.

hannyle (approved):

Issue #532, opened 2021-11-08
New navigation bar


blankdots:
#### Description

as part of move to new design styles #363 we need a new navigation bar that adheres to:
https://www.figma.com/file/NJiSfMKCPMJZIduAYVq0he/SD-vs-2?node-id=0:1

language is going be under the username

![image](https://user-images.githubusercontent.com/47524/142606705-649d1b6b-6d86-47a1-a3c5-911fc3600ee4.png)

#### DoD (Definition of Done)

New navigation bar

#### Testing

Peer review

hannyle:
A later improvement in design will have a new follow-up issue.

Issue #539, opened 2021-11-19
Redesign main page with drafts and published 


blankdots:
#### Description

Follow the template at https://www.figma.com/file/NJiSfMKCPMJZIduAYVq0he/SD-vs-2?node-id=0%3A1

![image](https://user-images.githubusercontent.com/47524/142654812-c0d8fdbf-184e-44a0-ae88-8ddc350cc262.png)

#### DoD (Definition of Done)

New main page with drafts and published folders/submissions

#### Testing

Unit and integration tests

hannyle:
![Screenshot 2022-02-18 at 14 57 32](https://user-images.githubusercontent.com/23532204/154687162-68ce185c-5192-49a1-81f2-c121415166a9.png)

PR #669, opened 2022-02-18
Feature/new-navigation-bar-and-home-page

hannyle:
### Description

Implementing new navigation bar and home page according to new UI design.

### Related issues

Closes https://github.com/CSCfi/metadata-submitter-frontend/issues/532 and https://github.com/CSCfi/metadata-submitter-frontend/issues/539

### Type of change

### Changes Made

- Moved Log out button to be under User profile in Navigation bar
- Moved all draft submissions and published submissions to Home page
- Separated draft submissions and published submissions in two different tabs
- Converted submission's `dateCreated` from unix timestamp to ISO8601
- Added functionality to sort the dates
- Updated pagination to show more details of page selection and rendered it responsively in smaller screen sizes
- Added breakpoints for font-sizes in different screen sizes
- Removed components and test files for deprecated features
- Update unit tests and integration tests for current UI state

### Testing

- [x] Unit Tests
- [x] Integration Tests

### Mentions

This is not a final work. Improvements and more detailed responsive views will be implemented in later rounds when the final UI design is made.


blankdots (changes requested):
Some small things, otherwise looks good to me.

i noticed these in Firefox 97 for Ubuntu

Peek 2022-02-21 12-35
this is a bit off, is there any chance we can fix the hover on the user menu ?
I suspect it is because of arrow up icon is smaller than the arrow down one.

--- 
Peek 2022-02-21 12-38
There seems to be a weird border on select for table elements, either we drop it or we add the border top as well, (assumption maybe the border top is there, but it is missing a z-index ?)

hannyle:
> Some small things, otherwise looks good to me.
> 
> i noticed these in Firefox 97 for Ubuntu
> 
> Peek 2022-02-21 12-35 this is a bit off, is there any chance we can fix the hover on the user menu ? I suspect it is because of arrow up icon is smaller than the arrow down one.

--> You're right. The icons were not the same size. This is fixed!
> 
> Peek 2022-02-21 12-38 There seems to be a weird border on select for table elements, either we drop it or we add the border top as well, (assumption maybe the border top is there, but it is missing a z-index ?)

--> This is a default thing of MUI DataGrid. I removed this weird border on selecting table elements.

blankdots (approved):

Issue #695, opened 2022-03-09
Study requires Abstract and Dataset requires Description


blankdots:
#### Description

As part of the changes from https://github.com/CSCfi/metadata-submitter/pull/356 study requires now `StudyAbstract` and dataset requires `Description`

We need to adapt the tests to fit changes

might require changes from: https://github.com/CSCfi/metadata-submitter-frontend/issues/628
#### DoD (Definition of Done)

Test pass again with changes from https://github.com/CSCfi/metadata-submitter/pull/356

#### Testing

Cypress tests

Issue #628, opened 2022-01-25
add list of keywords to DOI form


blankdots:
#### Description

with PR https://github.com/CSCfi/metadata-submitter/pull/332 in backend the `doi` schema now requires `keywords` as part of the form. 

Make use of https://mui.com/components/chips/#chip-array or https://csc-design-system-docs.web.app/tags to allow users to add a list of keywords/tags

#### DoD (Definition of Done)

keywords can be added as part of the DOI form

#### Testing

Cypress tests

blankdots:
changes here: https://github.com/CSCfi/metadata-submitter/pull/356

PR #715, opened 2022-03-21
Feature/DOI-keywords-and-integration-tests

hannyle:
### Description

- `Keywords` is a mandatory field in DOI form and users can type and save as many `keywords` as possible. Keywords are displayed in the form of tags so after typing some text, users should press `,` or `Enter` to mark the keyword as a tag.
-  Fix integration tests for `Study` object, now `Study` form has a mandatory `Study Abstract` field and we can only submit one `Study` object.

### Related issues

Solves https://github.com/CSCfi/metadata-submitter-frontend/issues/628 and https://github.com/CSCfi/metadata-submitter-frontend/issues/695

### Type of change

### Changes Made

- In `WizardJSONSchemaParser.tsx`, MUI Chip component was used to render `Keywords` field. A hidden `input` is used to register value to the form and MUI Textfield is used to receive input text from user.
- Updated Cypress tests to include:
  - `Study Abstraction` when filling a Study form and only one Study form is submitted per time.
  - `Dataset` object is mandatorily submitted before publishing a submission
  - `DOI` form is also saved before publishing a submission. A Cypress command `saveDoiForm` is added to be reused.

### Testing

- [x] Integration Tests

### DIFF
--- a/.github/workflows/e2etests.yml
+++ b/.github/workflows/e2etests.yml
@@ -22,7 +22,7 @@ jobs:
           if [[ $BRANCH == master ]]; then
              echo "VERSION=master" >> $GITHUB_ENV
           else
-             echo "VERSION=feature/metax-integration" >> $GITHUB_ENV
+             echo "VERSION=develop" >> $GITHUB_ENV
           fi
       - name: Clone backend
         uses: actions/checkout@v3
--- a/.github/workflows/e2etests.yml
+++ b/.github/workflows/e2etests.yml
@@ -22,7 +22,7 @@ jobs:
           if [[ $BRANCH == master ]]; then
              echo "VERSION=master" >> $GITHUB_ENV
           else
-             echo "VERSION=develop" >> $GITHUB_ENV
+             echo "VERSION=feature/metax-integration" >> $GITHUB_ENV
           fi
       - name: Clone backend
         uses: actions/checkout@v3
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Field Keywords in DOI form #715
 - Option to filter by submission's name in Home page #685
 - New navigation bar and new home page #669
 
--- a/cypress/integration/app.spec.ts
+++ b/cypress/integration/app.spec.ts
@@ -1,5 +1,3 @@
-import { FormInput } from "../support/types"
-
 describe("Basic e2e", function () {
   beforeEach(() => {
     cy.task("resetDb")
@@ -9,7 +7,7 @@ describe("Basic e2e", function () {
     cy.login()
   })
 
-  it("should create new folder, add Study form, upload Study XML file, add Analysis form, add DAC form, and publish folder", () => {
+  it("should create new folder, add Study form, add Analysis form, add DAC form, and publish folder", () => {
     cy.login()
 
     cy.get("button", { timeout: 10000 }).contains("Create submission").click()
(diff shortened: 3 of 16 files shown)


blankdots:
it would be nice to update changelog

hannyle:
I understand that we can only submit one `Study` object, but is it also forbidden to edit or replace a `Study` object? It seems so to me with the current backend changes.

blankdots:
> @blankdots I understand that we can only submit one `Study` object, but is it also forbidden to edit or replace a `Study` object? It seems so to me with the current backend changes.

are you referring to XML or form submissions?

hannyle:
> > @blankdots I understand that we can only submit one `Study` object, but is it also forbidden to edit or replace a `Study` object? It seems so to me with the current backend changes.
> 
> are you referring to XML or form submissions?
> 
> from what I see in integration tests in backend we are able to do `PATCH` on `Study` https://github.com/CSCfi/metadata-submitter/blob/feature/metax-integration/tests/integration/run_tests.py#L1770

For me it happens in both form and XML. Other objects work fine but with `Study` form, for example, editing and updating a submitted form caused error: `"Some items (e.g: "accessionId", "publishDate", "dateCreated", "metaxIdentifier") cannot be changed."`

blankdots:
> For me it happens in both form and XML. Other objects work fine but with `Study` form, for example, editing and updating a submitted form caused error: `"Some items (e.g: "accessionId", "publishDate", "dateCreated", "metaxIdentifier") cannot be changed."`

Yes, that is correct i think "accessionId", "publishDate", "dateCreated" was never possible to update any of them previously as well, it is good not to include them in the `PATCH` JSON. if the `PATCH` JSON does not include them, then we have a bug in backend. 
Could you provide the JSON and XML you are sending for PATCH update ?

Similar fields cannot be updated for XML as well. 
I think that is similar for all objects including datasets.

lilachic:
I'm not sure how keywords should be added, one at a field or can I but all the keywords in a field? 
Trying the DOI form manually  I got a bit weird error for 2nd and 3rd keywords, although there are some characters it says "must NOT be shorter than 1 character"
Screenshot 2022-03-22 at 15 18 09

hannyle:
> Yes, that is correct i think "accessionId", "publishDate", "dateCreated" was never possible to update any of them previously as well, it is good not to include them in the `PATCH` JSON. if the `PATCH` JSON does not include them, then we have a bug in backend. Could you provide the JSON and XML you are sending for PATCH update ?
> 
> Similar fields cannot be updated for XML as well. I think that is similar for all objects including datasets.

Yes, an example of updating a submitted `Study` form where I only changed the `title`:

POST request:
Screenshot 2022-03-22 at 15 39 16

PATCH request:
Screenshot 2022-03-22 at 15 34 33

As of now I have checked again, the error only happens with updating `Study` form, replacing the XML file works fine.

blankdots:
yes so you should not send `metaxIdentifier` and `doi` in `PATCH request that is why it fails -  well `metaxIdentifier` is now a forbidden key and probably we will make `doi` a forbidden key as well

hannyle:
> I'm not sure how keywords should be added, one at a field or can I but all the keywords in a field? Trying the DOI form manually I got a bit weird error for 2nd and 3rd keywords, although there are some characters it says "must NOT be shorter than 1 character" Screenshot 2022-03-22 at 15 18 09


Sorry for not making it clear in the description, `Keywords` are display in the form of `tag` so you need to press `,` or `Enter` after typing the keyword. I updated the description here: 

Screenshot 2022-03-22 at 15 47 22

blankdots:
> @hannyle yes, so you should not send `metaxIdentifier` and `doi` in `PATCH request that is why it fails - well `metaxIdentifier`is now a forbidden key and probably we will make`doi` a forbidden key as well

the same fields should be omitted for `dataset` object as well

hannyle:
> @hannyle yes, so you should not send `metaxIdentifier` and `doi` in `PATCH request that is why it fails - well `metaxIdentifier`is now a forbidden key and probably we will make`doi` a forbidden key as well


> > @hannyle yes, so you should not send `metaxIdentifier` and `doi` in `PATCH request that is why it fails - well `metaxIdentifier`is now a forbidden key and probably we will make`doi` a forbidden key as well
> 
> the same fields should be omitted for `dataset` object as well

You're right, `metaxIdentifier` and `doi` shouldn't be in the PATCH. I will modify the request in frontend.

saulipurhonen:
Not sure if this should be tackled in another PR but input field doesn't seem to adjust with the keywords
image

hannyle:
> Not sure if this should be tackled in another PR but input field doesn't seem to adjust with the keywords image

Good catch! I will fix it in this PR.

blankdots (approved):
The UI is work in progress and it shows, but that is understandable as we transition. The data looks good and seems to be registered correctly in the backend

saulipurhonen (approved):

saulipurhonen:
Looks good to me. Keywords behave as supposed and tests looks comprehensive.

Issue #790, opened 2022-05-03
DOI form should show catch error when Keywords field is not filled


hannyle:
#### Describe the bug

Saving DOI form without filling `Keywords` has redirected the page to `error400`. 

#### To Reproduce

Steps to reproduce the behaviour:

1. Go to DOI form
2. Fill all required fields except `Keywords`
3. Click Save DOI form
4. See `error400` page

#### Expected behavior

Error of not filling `Keywords` field should be caught in front-end and should not redirect to error400 page.

PR #792, opened 2022-05-04
Catch the error for not filling Keywords in DOI form

hannyle:
### Description

Saving DOI form without filling required `Keywords` field should show error in the form and not being redirected to error page 400.

### Related issues

Fixes https://github.com/CSCfi/metadata-submitter-frontend/issues/790

### Type of change

### Changes Made

- Modify `FormTagField` to catch error correctly when they Keyword field is not filled.

### Testing

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Tests do not apply

### DIFF
--- a/src/components/SubmissionWizard/WizardForms/WizardJSONSchemaParser.tsx
+++ b/src/components/SubmissionWizard/WizardForms/WizardJSONSchemaParser.tsx
@@ -1276,8 +1276,7 @@ const FormTagField = ({ name, label, required, description }: FormFieldBaseProps
             name={name}
             control={control}
             defaultValue={defaultValue}
-            rules={{ required: true }}
-            render={({ field, formState }) => {
+            render={({ field }) => {
               const handleKeywordAsTag = (keyword: string) => {
                 // newTags with unique values
                 const newTags = !tags.includes(keyword) ? [...tags, keyword] : tags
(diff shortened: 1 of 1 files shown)


genie9:
fixes #790

genie9 (approved):
Keywords cannot be empty anymore, which is great! The only thing which I wonder is the style issue. Keywords required field is looking different from other required fields.
Screenshot from 2022-05-09 15-41-49
It is implemented differently but is there a possibility that warning texts would look the same?

hannyle:
> Keywords cannot be empty anymore, which is great! The only thing which I wonder is the style issue. Keywords required field is looking different from other required fields. Screenshot from 2022-05-09 15-41-49 It is implemented differently but is there a possibility that warning texts would look the same?

Yes, It's good that you are pointing out the differences in showing error styles. 

The fact is that in DOI form,`Keywords` field is a text field (string type) while other required fields are nested items (array type). The errors are shown based on field's `type` and that's why they appears differently. This happens not only in DOI form but also in other forms.

We can modify them to look the same way (in another PR) but before that I think it's better to consult @ainoc for the design :)

saulipurhonen (approved):
Code looks clean and I'm not able to reproduce original problem :)

PR #811, opened 2022-05-24
Feature/form and xml changes

hannyle:
### Description

Introduced new structure of  **Form**, new way of showing **Tooltip** and new way to upload and show **XML File**.

### Related issues

Closes  https://github.com/CSCfi/metadata-submitter-frontend/issues/540, https://github.com/CSCfi/metadata-submitter-frontend/issues/541, https://github.com/CSCfi/metadata-submitter-frontend/issues/787

### Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Changes Made

- Converted the **Form** into new layout by mainly refactoring `FormSection`  in `WizardJSONSchemaParser.tsx`
- Remade **Tooltip** with reusable component `DisplayDescription`, limited the default description text length to 60 chars with option to **Read more/Expand** text
- Added **Options** including **Upload XML** and **Clear form** to **Form** 
- Changed form's action buttons to **Save as draft** and **Mark as ready**
- Added modal for selecting/drag-dropping XML file
- Added a new page for showing information of selected XML file
- Fixed a bug about refreshing and starting a new form whenever we click `Add <object>` button
- Update unit test for **Tooltip** 
- Update e2e tests for **Form**

### Testing

- [x] Unit Tests
- [x] Integration Tests

PR #835, opened 2022-06-14
Feature/form and xml changes

hannyle:
### Description

This is the same PR as https://github.com/CSCfi/metadata-submitter-frontend/pull/811 but with additional modification for github actions files.

### Changes Made

- After trying different ways of fixing the github workflow files, I think the safe option in long run would be recreating the `package-lock.json` file and keep the `npm ci` as it is in those files.
- Modified the backend branch in `e2etests.yml` from `develop` to SHA `e8f9ec4181f5c9b03c6f092b423652316e60c418`


lilachic (approved):
LGTM. If this induces technical dept do we have issue to solve that later?

PR #874, opened 2022-09-29
Merge master into develop

csc-felipe:
Let's fix the dissonance between `master` and `develop`!

Like in https://github.com/CSCfi/metadata-submitter/pull/595


blankdots (approved):

teemukataja (approved):

hannyle:
Just to make sure I understand correctly, are we deprecating the `master` branch and use `develop` as the main branch ?

csc-felipe:
> Just to make sure I understand correctly, are we deprecating the `master` branch and use `develop` as the main branch ?

I had proposed that we would change so that `develop` would be the `default` branch, so that, for example, changing GitHub issue templates would be done in the `develop` branch instead of `master`.

I didn't propose to remove the `master` branch, but it could for e.g. be renamed into `releases`.

This PR was so that it's possible to create a PR from `develop` into `master` again, because at the moment there are conflicts, and merging `master` to `develop` fixes it.

We can discuss again if we change the branches or not. We can anyway still merge this PR.

hannyle (approved):

